PR #8, opened 2017-02-21
Add Travis-CI build and testing

afonso360:
Hi,

This PR add Contious integration from travis-ci.
It tests stable, beta and nightly on linux.

It doesen't test on osx, because ive had some bad experiences adding osx to travis and the builds suddenly starting to take 3 hours.

fast_finish means that if other jobs fail it will mark the build completed immediately instead of waiting for it to finish.

I could also try to add appveyor ci if you would like me to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: rust
+rust:
+  - stable
+  - beta
+  - nightly
+os:
+  - linux
+
+matrix:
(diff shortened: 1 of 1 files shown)


Manishearth:
This doesn't really add any tests, but I guess we can have the compile check for now and look into testing the behavior later.

PR #12, opened 2017-02-21
Rename --target to --script

nagisa:
--target is a common flag for other `cargo` subcommands and may be confusing, so
rename it to --script instead. This also opens up ability to eventually use
--target the same way other cargo subcommands do (e.g. to cross-compile fuzzers?)

r? @Manishearth


frewsxcv:
Why not `--fuzz-target`? 'target' is a very common term when talking about fuzzing, I'd be hesitant to stray away from it.

http://llvm.org/docs/LibFuzzer.html#fuzz-target

nagisa:
fuzz-target seems fine to me, will file a followup PR.

PR #22, opened 2017-02-22
Don't overwrite user-defined RUSTFLAGS

### LINKED ISSUE #9: RUSTFLAGS are ignored
It seems like `RUSTFLAGS` are ignored when compiling the fuzzer script. This makes it impossible to fuzz crates that rely on e.g. AVX2.

vks:
Fixes #9 by appending to RUSTFLAGS instead of overwriting it.

I was not sure how to test this, so the code is untested unfortunately. I only made sure it compiles.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -180,6 +180,11 @@ fn rebuild_libfuzzer() -> io::Result<()> {
 fn run_target(target: String) -> io::Result<bool> {
     env::set_current_dir("./fuzz")?;
     rebuild_libfuzzer()?;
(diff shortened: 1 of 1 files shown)


Manishearth:
You can run `cargo install -f` in the cargo-fuzz folder to overwrite your installed cargo fuzz and test it.

Manishearth:
looks good to me though. let me know how it works after you test it.

vks:
Yes, it seems to work!

PR #26, opened 2017-02-23
Use a corpus

Manishearth:
r? @frewsxcv

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -103,6 +103,7 @@ path = "fuzzers/fuzzer_script_1.rs"
 write!(ignore, r#"
 target
 libfuzzer
+corpus
 "#)?;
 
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -212,9 +213,12 @@ fn run_target(target: String) -> io::Result<bool> {
         return Err(io::Error::new(io::ErrorKind::Other, "Failed to build fuzz target"))
     }
 
+    let _ = fs::create_dir("corpus");
+
     // can't use cargo run since we can't pass -L args to it
     let path = format!("target/debug/{}", target);
     let mut run_cmd = process::Command::new(path);
+    run_cmd.arg("corpus");
```
nagisa: Shouldn’t pass the argument if creating the directory failed for any reason other than the directory already existing.

Manishearth: I made it bail if creating the dir fails.

PR #32, opened 2017-02-24
Set detect_odr_violations=0

### LINKED ISSUE #7: Run with detect_odr_violations=0
Probably should turn off most of the asan checks. I needed `ASAN_OPTIONS=detect_odr_violation=0` to make it work on rust-url, otherwise it would bail very early.

Manishearth:
r? @nagisa
(fix #7)


nagisa:
r=me

PR #33, opened 2017-02-24
Support artifact directory

### LINKED ISSUE #6: Set up a corpus and artefact directory
Currently the artefacts are just dumped into the fuzz folder. We should create a corpus dir and pass it to the binary. We should also set up the `-artefact-prefix` arg

http://llvm.org/docs/LibFuzzer.html#options

Manishearth:
(fix #6)

r? @frewsxcv @nagisa

(contains code from other PR since it would be confliciting otherwise)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -242,7 +242,8 @@ fn run_target(target: String) -> Result<bool, Box<error::Error>> {
     // can't use cargo run since we can't pass -L args to it
     let path = format!("target/debug/{}", target);
     let mut run_cmd = process::Command::new(path);
-    run_cmd.arg("corpus");
+    run_cmd.arg("corpus")
+           .env("ASAN_OPTIONS", "detect_odr_violation=0");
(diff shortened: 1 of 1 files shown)


frewsxcv:
Seems alright to me. Will wait for @nagisa though

nagisa:
LGTM too.

Manishearth:
Hold on before merging, need to test and tweak.

nagisa:
(again, just like with corpus, we should have a per-fuzz-target subdirectory for these)

Manishearth:
> (again, just like with corpus, we should have a per-fuzz-target subdirectory for these)

I'm not sure if this is necessarily true. Might be better if stuff was shared for the corpus. For artifacts I guess it makes sense. Would be better if done when we figure out how to make libfuzzer run indefinitely, it currently doesn't, so you only have one file to worry about anyway and you can see that in stdlout.

Issue #2, opened 2017-02-21
Better error handling


Manishearth:
Add colors, use `error_chain`, etc.

ner0x652:
@Manishearth I can help with colors part(I have, a little, experience with `term` [crate](https://crates.io/crates/term). On `error_chain` I will need some more time.

Manishearth:
Go ahead!

ner0x652:
@Manishearth assign it to me please, I'll work on it during the weekend.

Manishearth:
Can't assign out of org. Take the above comment  as indication that it's assigned

ner0x652:
@Manishearth by org you mean rust-fuzz ? If yes, then I'm part of it...(even if a tiny one :smile:)

Manishearth:
Oh, I didn't realize that. Done.

nagisa:
error_chain-ed errors have been added in some places.

PR #35, opened 2017-02-26
Added (partial) terminal coloring. Issue #2

ner0x652:


Cargo.toml
```diff
@@ -1,12 +1,13 @@
...
 description = "Simple wrapper around libFuzzer"
 repository = "https://github.com/rust-fuzz/cargo-fuzz/"
 
 [dependencies]
-docopt = "0.7"
+docopt          = "0.7"
 rustc-serialize = "0.3"  # if you're using `derive(RustcDecodable)`
-cargo_metadata = "0.1.2"
+cargo_metadata  = "0.1.2"
+term            = "0.4.5"
```
Manishearth: newline at EOF

src/main.rs
```diff
@@ -56,34 +59,63 @@ fn main() {
...
         }
     } else if args.flag_list {
-        list_fuzz_targets()
+        list_fuzz_targets(&mut term_stdout)
             .map(|_| ())
     } else {
-        println!("Invalid arguments. Usage:\n{}", USAGE);
+        if let Some(ref mut terminal) = term_stdout {
+            let _ = terminal.fg(term::color::GREEN);
+            println!("Invalid arguments. Usage:\n{}", USAGE);
```
Manishearth: These need to be written to the term stdout/err (probably stderr)

Also, can we follow the cargo/rustc convention here? They show a white bold "error: some error message" where `error` is red. So we would display one bold/colorful line, followed by the usage in the default color.

src/main.rs
```diff
@@ -56,34 +59,63 @@ fn main() {
...
+        }
+
         return;
     };
     if let Err(error) = result {
-        writeln!(io::stderr(), "Error: {}", error)
-            .expect("failed writing to stderr");
+        if let Some(ref mut terminal) = term_stderr {
+            let _ = terminal.fg(term::color::RED);
```
Manishearth: Same thing here, `Error:` should be red and bold, whereas the message should be white and bold.

ner0x652:
@Manishearth please review the latest commit

src/main.rs
```diff
@@ -56,34 +59,73 @@ fn main() {
...
                 Ok(())
             } else {
-                println!("Fuzzing found errors!");
+                if let Some(ref mut terminal) = term_stdout {
+                    let _ = terminal.fg(term::color::YELLOW);
+                    println!("Fuzzing found errors!");
+                    let _ = terminal.reset();
+                } else {
+                    println!("Fuzzing found errors!");
+                }
```
frewsxcv: Can we move all this logic into a separate function now that it's getting a little noise? Something like:

```rust
fn print_found_errors() {
  let mut term_stdout = term::stdout();
  let mut term_stderr = term::stderr();
  if let Some(ref mut terminal) = term_stdout {
      let _ = terminal.fg(term::color::YELLOW);
      println!("Fuzzing found errors!");
      let _ = terminal.reset();
    } else {
      println!("Fuzzing found errors!");
    }
  }
}
```

frewsxcv: (same with the other places there's `term` logic)

ner0x652: yes we can. I suggest to move it to a separate module, like `utils`, where we can add other logic like the one from `term`

ner0x652: @frewsxcv @Manishearth @nagisa So, are you ok with adding the `term` code to a new module, say `utils` or leaving it into `lib` ?

frewsxcv: Either works for me

ner0x652:
@Manishearth @nagisa @frewsxcv When you have some time, please review my latest commit

src/main.rs
```diff
@@ -59,71 +58,33 @@ fn main() {
...
-            let _ = terminal.fg(term::color::WHITE);
-            writeln!(io::stderr(), " Invalid arguments. Usage:\n{}", USAGE)
-                .expect("failed writing to stderr");
-            let _ = terminal.reset();
-        } else {
-            writeln!(io::stderr(), "Invalid arguments. Usage:\n{}", USAGE)
-                .expect("failed writing to stderr");
-        }
-
+        utils::write_to_stderr(format!("{}{}", "Invalid arguments. Usage:\n{}", USAGE).as_str());
```
Manishearth: This won't do the right thing. You want `format!("Invalid arguments. Usage:\n{}", USAGE)`.

ner0x652:
Thank you for helpful input and patience! :smile:

src/main.rs
```diff
@@ -66,24 +69,25 @@ fn main() {
         list_fuzz_targets()
             .map(|_| ())
     } else {
-        println!("Invalid arguments. Usage:\n{}", USAGE);
+        utils::write_to_stderr(format!("Invalid arguments. Usage:\n{}", USAGE).as_str());
```
Manishearth: Actually, the usage string should not be bold, so you should use a different call to do this.

ner0x652: Ok, I'll change it.

Manishearth:
Answering your question in IRC, I meant something like

`fn write_to_strerr(msg: &str, extended: Option<&str>)`. If `extended` is None, it prints a bold `error: <msg>` (with the right colors). If it is `Some`, it prints a bold `error: <msg>`, and then a non-bold non-colored `<extended>` on the next line. Then for the usage stuff we call it as `write_to_stderr("Invalid arguments. Usage:", USAGE)`

ner0x652:
Ok, now I've got it.

În Mar, 28 feb. 2017, 00:12 Manish Goregaokar, <notifications@github.com> a
scris:

Manishearth:
nevermind, github lets me squash directly

PR #38, opened 2017-02-27
Transition from docopt to clap

### LINKED ISSUE #37: Move to clap
@nagisa has mentioned this a few times

blt:
This commit moves cargo-fuzz away from docopt to the more explicit
use of clap. This resolves #37.

Signed-off-by: Brian L. Troutwine <brian@troutwine.us>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,50 +6,44 @@
 // copied, modified, or distributed except according to those terms.
 
 extern crate cargo_metadata;
-extern crate docopt;
+extern crate clap;
 extern crate rustc_serialize;
 
 use cargo_metadata::{metadata, Package};
-use docopt::Docopt;
-use std::{env, error, fs, io, path, process};
+use clap::{App, Arg};
+use std::{convert, env, error, fmt, fs, io, path, process};
 use std::io::Write;
 
-const USAGE: &'static str = "
-Cargo Fuzz
-
-Usage:
-  cargo fuzz --init
-  cargo fuzz --fuzz-target TARGET
-  cargo fuzz --add TARGET
-  cargo fuzz --list
-  cargo fuzz (-h | --help)
-
-Options:
-  -h --help              Show this screen.
-  --init                 Initialize fuzz folder
-  --fuzz-target TARGET   Run with given fuzz target in fuzz/fuzzers
-  --add TARGET           Add a new fuzz target
-  --list                 List the available fuzz targets
-";
-
-#[derive(Debug, RustcDecodable)]
-struct Args {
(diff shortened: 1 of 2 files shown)


frewsxcv:
cc @ner0x652

src/main.rs
```diff
@@ -62,11 +56,12 @@ fn main() {
         } else {
             result.map(|_| ())
         }
-    } else if args.flag_list {
+    } else if args.is_present("list") {
         list_fuzz_targets()
             .map(|_| ())
     } else {
-        println!("Invalid arguments. Usage:\n{}", USAGE);
+        println!("Invalid arguments. Usage:\n{}",
```
nagisa: Should just let clap to print out the help message if parsing fails.

src/main.rs
```diff
@@ -6,50 +6,44 @@
...
+            .help("Add a new fuzz target")
+            .takes_value(true))
+        .arg(Arg::with_name("list")
+            .long("list")
+             .help("List the available fuzz targets"));
+    let mut app_help = Vec::new();
+    app.write_help(&mut app_help).ok().expect("could not write help");
+    let args = app.get_matches();
+
+    let result = if args.is_present("init") {
```
nagisa: Should *either* refactor this a bit so something like `cargo fuzz --init --add=name` works (i.e. not if-else chain) or change the CLI a bit to either make the flags exclusive or make the flags subcommands (its fine to break). 

I’d personally prefer changing it, so we’d have something like:

```
cargo fuzz init
cargo fuzz add NAME --maybe --some --options
cargo fuzz target NAME --some --rustc --options -- fuzz options
```

src/main.rs
```diff
@@ -159,7 +154,7 @@ pub extern fn go(data: &[u8]) {{
 }
 
 /// Add a new fuzz target script with a given name
-fn add_target(target: String) -> Result<(), Box<error::Error>> {
+fn add_target<S>(target: S) -> Result<(), Box<error::Error>> where S: Into<String> + fmt::Display {
```
nagisa: It is probably just better to take string and convert in the caller.

src/main.rs
blt: I can get behind that, sure.

src/main.rs
blt: Ah yeah, it does that already. I added this bit to keep existing behaviour but I'll happily drop it.

src/main.rs
blt: Heard.

nagisa:
This has been superseded. By other PRs.

Issue #44, opened 2017-02-28
Entity not found


ner0x652:
Running `cargo fuzz --fuzz-target fuzzer_script_1` results in a `Error: Entity not found`.

Discussed with @nagisa about this issue on IRC, the cause is
> "since we started passing in the `--target=xxx` flag, it puts the executable in `fuzz/target/<TARGET>/debug` where cargo-fuzz looks for it in `fuzz/target/debug`"

nagisa:
After https://github.com/rust-fuzz/cargo-fuzz/pull/51 this should be fixed.

Issue #29, opened 2017-02-24
Consider `cargo fuzz <fuzz target>` instead of `cargo fuzz --fuzz-target <fuzz target>`


frewsxcv:
Considering `cargo fuzz --fuzz-target <fuzz target>` is the main command here and going to be called the most, it seems unfortunate it's a pretty long CLI call. Maybe we should migrate to `cargo fuzz <fuzz target>`? Thoughts?

nagisa:
I was thinking of doing that “implicitly” with a move to clap.

Manishearth:
(I'm totally open to a move to clap or any other parser, I just picked docopt because cargo uses it. I'm also not attached to the current argument format. Go ahead and do whatever 😄 )

nagisa:
This is fixed in #47, I think?

It is now `cargo fuzz run target`

Issue #3, opened 2017-02-21
More configurability


Manishearth:
Add the ability to choose between debug/release. Pass down any other useful rustc flags.

ruuda:
Being able to pass arguments to libfuzzer, such as `-max_size=` or the location of the corpus, would also be useful. Currently I run `cargo fuzz --fuzz-target fuzzer_script_1` to trigger the build, kill it, and then I start `fuzz/target/debug/fuzzer_script_1` with the desired flags.

Simply passing on any extra arguments to the target program would be sufficient.

Manishearth:
Yep. `cargo fuzz --fuzz-target foo -- FUZZ ARGS BLAH BLAH` is what we'd probably use.

nagisa:
You can pass args to the underlying binary now

(try `cargo fuzz run target -- -help=1`)

It is still desirable to have more configuration for how the fuzz target is compiled (release/debug, debug-assertions on or off, etc)

Bobo1239:
Just asking for confirmation: Ignoring a function/crate (like https://clang.llvm.org/docs/SanitizerSpecialCaseList.html) would need rustc support, right?
(In my case [rayon](https://github.com/nikomatsakis/rayon) is triggering ASan though I believe that is https://github.com/rust-lang/rust/issues/39882.)

nagisa:
No, not really. It likely can be expressed through `-C llvm-args` in `RUSTFLAGS`.

Bobo1239:
I did try `RUSTFLAGS="-C llvm-args=-fsanitize-blacklist=blacklist.txt" cargo fuzz run fuzzer_script_1` but that only results in
```
error: failed to run `rustc` to learn about target-specific information
Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names -C llvm-args=-fsanitize-blacklist=blacklist.txt -Cpasses=sancov -Cllvm-args=-sanitizer-coverage-level=3 -Zsanitizer=address -Cpanic=abort --crate-type bin --crate-type dylib --crate-type proc-macro --crate-type rlib --target x86_64-unknown-linux-gnu` (exit code: 1)
--- stderr
rustc: Unknown command line argument '-fsanitize-blacklist=blacklist.txt'.  Try: 'rustc -help'
rustc: Did you mean '-dfsan-abilist=blacklist.txt'?
```
[...]

PR #47, opened 2017-02-28
⚑⚐⚑⚐⚑⚐

nagisa:
Supersedes https://github.com/rust-fuzz/cargo-fuzz/pull/38
Fixes https://github.com/rust-fuzz/cargo-fuzz/issues/44
Fixes https://github.com/rust-fuzz/cargo-fuzz/issues/37
Fixes https://github.com/rust-fuzz/cargo-fuzz/issues/29
cc https://github.com/rust-fuzz/cargo-fuzz/issues/3 (allows passing arbitrary flags to invoked process)

This is what the CLI looks now:

```
$ cargo fuzz -h
cargo-fuzz 0.2.2
Simple wrapper around libFuzzer

USAGE:
    cargo-fuzz [dummy] [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

ARGS:


SUBCOMMANDS:
    add     Add a new fuzz target
    help    Prints this message or the help of the given subcommand(s)
    init    Initialize the fuzz folder
    list    List all fuzz targets
    run     Run the fuzz target in fuzz/fuzzers


$ cargo fuzz run fuzzer_script_1 -- -help=1
Usage:

To run fuzzing pass 0 or more directories.
target/x86_64-unknown-linux-gnu/debug/fuzzer_script_1 [-flag1=val1 [-flag2=val2 ...] ] [dir1 [dir2 ...] ]

To run individual tests without fuzzing pass 1 or more files:
target/x86_64-unknown-linux-gnu/debug/fuzzer_script_1 [-flag1=val1 [-flag2=val2 ...] ] file1 [file2 ...]

Flags: (strictly in form -flag=value)
 verbosity                   	1	Verbosity level.
 seed                        	0	Random seed. If 0, seed is generated.
 runs                        	-1	Number of individual test runs (-1 for infinite runs).
...
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,53 +6,47 @@
 // copied, modified, or distributed except according to those terms.
 
 extern crate cargo_metadata;
-extern crate docopt;
+extern crate clap;
 extern crate rustc_serialize;
 extern crate term;
 
 use cargo_metadata::{metadata, Package};
-use docopt::Docopt;
-use std::{env, error, fs, io, path, process};
+use clap::{App, Arg};
+use std::{convert, env, error, fmt, fs, io, path, process};
 use std::io::Write;
 
 mod utils;
 
-const USAGE: &'static str = "
-Cargo Fuzz
-
-Usage:
-  cargo fuzz --init
-  cargo fuzz --fuzz-target TARGET
-  cargo fuzz --add TARGET
-  cargo fuzz --list
-  cargo fuzz (-h | --help)
-
-Options:
-  -h --help              Show this screen.
-  --init                 Initialize fuzz folder
-  --fuzz-target TARGET   Run with given fuzz target in fuzz/fuzzers
-  --add TARGET           Add a new fuzz target
-  --list                 List the available fuzz targets
-";
-
-#[derive(Debug, RustcDecodable)]
-struct Args {
-    flag_init: bool,
-    flag_add: Option<String>,
-    flag_fuzz_target: Option<String>,
-    flag_list: bool,
-}
+const VERSION: Option<&'static str> = option_env!("CARGO_PKG_VERSION");
 
 fn main() {
-    let args: Args = Docopt::new(USAGE)
-                            .and_then(|d| d.decode())
-                            .unwrap_or_else(|e| e.exit());
+    let app = App::new("cargo-fuzz")
+        .version(VERSION.unwrap_or("unknown"))
+        .arg(Arg::with_name("init")
+            .long("init")
+            .help("Initialize fuzz folder"))
+        .arg(Arg::with_name("fuzz-target")
+            .long("fuzz-target")
+            .value_name("TARGET")
+            .help("Run with given fuzz target in fuzz/fuzzers")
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -223,7 +197,13 @@ fn make_dir_if_not_exist(dir: &str) -> Result<(), io::Error> {
     Ok(())
 }
 /// Fuzz a given fuzz target
-fn run_target(target: String) -> Result<bool, Box<error::Error>> {
+fn exec_target<'a>(args: &ArgMatches<'a>) -> Result<(), Box<error::Error>> {
+    let target: String = args.value_of_os("TARGET").expect("TARGET is required").to_os_string()
+        .into_string().map_err(|_| "TARGET must be valid unicode")?;
+    let exec_args = args.values_of_os("ARGS");
+
+    let target_triple = "x86_64-unknown-linux-gnu";
```
Manishearth: For now can we assert if target is not target_triple?

Manishearth:
Also, does this need a readme update?

src/main.rs
```diff
@@ -6,89 +6,47 @@
...
-
-    env::set_current_dir("./fuzz")?;
-    let package = get_package();
-    for target in &package.targets {
-        utils::print_message(target.name.as_str(), term::color::GREEN);
-    }
-
-    Ok(())
+    let app = App::new("cargo-fuzz")
+        .version(option_env!("CARGO_PKG_VERSION").unwrap_or("0.0.0"))
```
frewsxcv: There's also https://docs.rs/clap/2.20.5/clap/macro.crate_version.html

src/main.rs
```diff
@@ -6,89 +6,47 @@
...
-    } else if let Some(target) = args.flag_add {
-        add_target(target)
-    } else if let Some(target) = args.flag_fuzz_target {
-        let result = run_target(target);
-        if let Ok(success) = result {
-            if success {
-                // Can this ever happen?
-                Ok(())
-            } else {
-                utils::print_message("Fuzzing found errors!", term::color::YELLOW);
```
frewsxcv: Where did this printing go?

nagisa: The fuzzer binary is now `exec`d. Moreover I’m not sure about proclaiming "fuzzing" found errors based purely on the fact that binary exit with non-0 code too. We’ve observed at least multiple cases where this could be true in practice:

* OOM (not necessarily an error uncovered by fuzzing);
* Bugs in the fuzzer itself.

So instead of claiming something, I just made it claim nothing, also getting rid of some small over-head in process.

src/main.rs
nagisa: Unwraps.

src/main.rs
nagisa: Hmm? `target` is the script to be fuzzed, not target triple. I also can see somebody naming their fuzz script a `x86_64-unknown-linux-gnu` and there’s no technical reason to forbid it?

Manishearth: oh, blegh.

PR #54, opened 2017-03-04
Do not send email notifications for failed builds

nagisa:
Since the repo is now managed by bors and failures will not land(?) to master
these emails provide no benefit whatsoever.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,5 @@ os:
 
 matrix:
   fast_finish: true
+notifications:
+  email: false


Manishearth:
We need bors permissions though

frewsxcv:
Asking @notriddle about the bors permission thing on IRC right now. I _think_ you can get around it by publicizing your membership to the rust-fuzz org.

notriddle:
Bors-NG doesn't care about GitHub perms. Add reviews by going to the bors dashboard and going into the repo, then into Settings.

frewsxcv:
Gotcha, will do that

frewsxcv:
For the record, I was able to add 'nagisa', but when I try 'manishearth' I get a 500. Thanks manish, always causing problems.

frewsxcv:
The error appears to be related to case sensitivity. I was able to add 'Manishearth'.

PR #62, opened 2017-03-05
Add vobsub bugs to trophy case

emk:
cargo-fuzz found 5 errors (so far!) in my MPEG-2 subtitle decoder library. Of these, three involved invalid slices, which suggests that I should replace all calls to `slice[x..y]` with a function call that returns an error.


frewsxcv:
Thanks!

bors r+

PR #65, opened 2017-03-05
Fix build for non unix platforms

aochagavia:


aochagavia:
I am not sure this is the most idiomatic way to deal with this error, since I have never used error_chain before. In case there is a better way please tell me so I can update my code.

By the way, I think this project needs an AppVeyor build to make sure the build on Windows doesn't accidentally break.

nagisa:
bors r+

---

Not sure if windows support matters much, when it simply doesn’t work there anyway.

aochagavia:
I didn't know that... I stumbled upon a blog post by @Manishearth and it looked dead simple so I decided to give it a try. I guess we should at least mention in the readme that only Unix platforms are supported.

Manishearth:
We do.

PR #75, opened 2017-03-06
Options to enable release and debug assertions

nagisa:


src/main.rs
```diff
@@ -142,30 +152,46 @@ impl FuzzProject {
...
+            "-Cpasses=sancov",
+            "-Cllvm-args=-sanitizer-coverage-level=3",
+            "-Zsanitizer=address",
+            "-Cpanic=abort",
+        ];
+        flags.extend(other_flags.split(" "));
+        let rustflags = if assertions {
+            &flags[..]
+        } else {
+            &flags[1..]
```
frewsxcv: This slicing logic seems like an easy way for a bug to creep up in the future. Why not just `insert` or `extend` the `Vec` instead?

frewsxcv:
FYI, Travis found an issue here

src/main.rs
```diff
@@ -39,9 +39,19 @@ fn main() {
...
             .setting(AppSettings::TrailingVarArg)
-            .arg(Arg::with_name("TARGET").required(true)
+            .arg(Arg::with_name("release")
+                 .long("release")
+                 .short("O")
+                 .help("Build artifacts in release mode, with optimizations"))
+            .arg(Arg::with_name("debug_assertions")
+                 .long("debug-assertions")
+                 .short("a")
+                 .help("Build artifacts with debug assertions enabled (default if not -O)"))
```
frewsxcv: Is it possible to specify both -a and -O?

nagisa: Sure. It’s an option.

Manishearth:
r+

nagisa:
bors r=Manishearth

PR #77, opened 2017-03-06
Nice stuff 😉

nagisa:


Manishearth:
I feel like it makes more sense to allow overriding corpus as a cargo fuzz arg, instead of splicing the extra args. Those should be passed down untouched.

nagisa:
This is somewhat difficult to achieve, because CLI like this:

```
cargo-fuzz run <TARGET> [<FILE>...] -- [<ARGS>...]
```

is not expressible with clap, sadly and

```
cargo-fuzz run <TARGET> [<FILE>...] [<ARGS>...]
```

is obviously ambiguous. We could have a flag that looks like `--no-default-corpus` but I feel like looking at the ARGS to figure if we should pass through our own corpus is trivially enough correct.

Manishearth:
You should have args before target.

This is a pretty common CLI pattern. `cmd <args> <name> -- <...>`

nagisa:
Why are we even travis-ing stable rust?

nagisa:
This is what the CLI looks like now, which is exactly what I was going for

```
    cargo-fuzz run <TARGET> [CORPUS]... [-- <ARGS>...]
```

Couldn’t be any better.

---

> You should have args before target.

That requires either copying the whole set of libfuzzer arguments into our CLI *or* restricting ourselves from adding our own arguments in the future.

Manishearth:
r+

nagisa:
r=Manishearth

nagisa:
bors r=Manishearth

---

you understand this instead maybe?

PR #79, opened 2017-03-06
Add script for travis CI test, Fixes #69

### LINKED ISSUE #69: Add proper tests for CI
It should cd to a test crate directory, that contains a single function which panics on an input of size 6 or something, then fuzz init, edit the fuzz target to call this function, run the fuzz target, and ensure a crashfile is produced of the correct length.

ner0x652:


.travis.yml
```diff
@@ -1,12 +1,21 @@
...
 os:
   - linux
 
 matrix:
   fast_finish: true
+
 notifications:
   email: false
+
+deploy:
```
frewsxcv: Can we just make this:

```
script:
  scripts/deploy.sh
```

and maybe just rename `scripts/deploy.sh` to `test/run.sh`?
  
```

Comment (frewsxcv) [inline .travis.yml]: Could also put `testcrate` in the `test/` directory

Comment (ner0x652) [inline .travis.yml]: Rename `etc` to `test`, right?

Comment (ner0x652) [inline .travis.yml]: Ok, `run` is more appropriate.

**Comment by ner0x652** [comment]
@frewsxcv @Manishearth I've renamed the directories and the `bash` script. Please review.

File: .travis.yml (hunk 1 above)
Comment (Manishearth) [inline .travis.yml]: Yeah, this has nothing to do with deploying, `script` is the way to go.

File: test/testcrate/Cargo.toml [hunk 2]
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "testcrate"
+version = "0.1.0"
+authors = ["The rust-fuzz Project Developers"]
+
```
Manishearth: `publish=false`

.travis.yml
Manishearth: There should be a `cargo build` before the script is run, too

test/run.sh
```diff
@@ -0,0 +1,8 @@
+cd testcrate
```
Manishearth: needs to cargo install before cding

ner0x652: What it should install?

Manishearth: cargo-fuzz. Just `cargo install` in the cargo-fuzz root dir is enough.

ner0x652:
@Manishearth Updated.

.travis.yml
frewsxcv: There shouldn't be a `deploy` section here, it should just be `script`. Here's some more information:

https://docs.travis-ci.com/user/customizing-the-build#Customizing-the-Build-Step

test/run.sh
```diff
@@ -0,0 +1,8 @@
+cd ..
```
Manishearth: As far as I can tell the script is run from the root anyway. Please check the travis CI results.

Manishearth:
You need to add the bin directory to the PATH (`~/.cargo/bin` I think, but could be wrong)

Current one is causing failures on travis:

```
test/run.sh: line 4: cargo-fuzz: command not found

sed: can't read fuzz/fuzzers/fuzzer_script_1.rs: No such file or directory

test/run.sh: line 6: cargo-fuzz: command not found
```

also, this will only work on nightly, you should probably restrict this to that platform

ner0x652:
@Manishearth I did add the path to binaries to PATH. Now there is another problem, seems related to this https://github.com/rust-lang/cargo/issues/3192

Manishearth:
Add

```toml
[workspace]
members = ["."]
```

to the testcrate

ner0x652:
@Manishearth unfortunately that doesn't solve the problem
```
error: multiple workspace roots found in the same workspace:
  /home/travis/build/rust-fuzz/cargo-fuzz/test/testcrate/fuzz
  /home/travis/build/rust-fuzz/cargo-fuzz/test/testcrate
```

Manishearth:
Add a similar key to cargo-fuzz itself?

ner0x652:
@Manishearth no, every Cargo.toml has a workspace member, but the issue remains.

Manishearth:
I guess a workaround would be to copy this folder to the parent directory, cd there, and run the tests from there.

ner0x652:
Seems that moving the `tescrate` to root of `cargo-fuzz` still doesn't solve the issue. Probably we should wait for a solution from Cargo team. :thinking:

neosilky:
Can this get revisited? The issue posted to Cargo has been closed with the consensus that nightly works okay, which is fine for us.

Manishearth:
triggered a rebuild, let's see if it works

Manishearth:
Moved to  https://github.com/rust-fuzz/cargo-fuzz/pull/92

PR #84, opened 2017-03-12
Merge ASAN_OPTIONS instead of overwriting them

ruuda:
This concatenates the `detect_odr_violation=0` option to existing `ASAN_OPTIONS`, instead of overwriting them, which allows users to pass their own `ASAN_OPTIONS`.

One use case for this is limiting the fuzzing time by passing `-max_total_time` to libfuzzer: when the allowed time expires, the fuzzer will stop in the middle of a fuzzing target, and leak sanitizer will report an error, and exit with a nonzero exit code. Setting `ASAN_OPTIONS="detect_leaks=0"` works around this problem, but then `cargo-fuzz` must not overwrite those options. Concatenate them instead.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -158,6 +158,14 @@ impl FuzzProject {
         fs::create_dir_all("corpus")?;
         fs::create_dir_all("artifacts")?;
 
+        // Merge the asan options, so users can still provide their own options
+        // to e.g. disable the leak sanitizer. Options are colon-separated.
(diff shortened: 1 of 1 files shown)


ruuda:
Build failure appears to be caused by https://github.com/rust-lang/cargo/issues/3819.

frewsxcv:
Thanks!

bors r+

PR #96, opened 2017-03-30
Update README on how to check specific inputs

### LINKED ISSUE #60: add check artifact command
If the program panics an artifact is written. I then try to fix the bug causing the panic. Now I want to check if I really fixed the bug. A command like 
`cargo fuzz check my_fuzzer crash-2c66bb9f0dabeaf158f68a7eea3b5033080173c3` would be useful that invokes `go(data)` with this artifact again.

neosilky:
Ultimately, this should fix #60.

There is a caveat here in that libfuzzer errors out saying it is only running the single input and won't be fuzzing and this is detected as a memleak (with a long backtrace). I can't find an option to say "*it's okay to just take this input and run it*". Might look at `libfuzzer-sys` and see what's up.

Still, this documentation is okay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ $ cargo fuzz run fuzzer_script_1 # or whatever the target is named
 Then, wait till it finds something! More complex invocations are available as well. Consider
 looking at `cargo fuzz --help`, `cargo fuzz run --help` and others.
 
+Once you have found something and believe you have fixed it, re-run the fuzzer with the input:
+
+```sh
+$ cargo fuzz run fuzzer_script_1 fuzz/artifacts/fuzzer_script_1/<file mentioned in crash output>
+```
+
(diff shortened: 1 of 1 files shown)


neosilky:
Okay so there is a caveat here in that libfuzzer errors out saying it is only running the single input and won't be fuzzing and this is detected as a memleak (with a long backtrace). I can't find an option to say "*it's okay to just take this input and run it*". Might look at `libfuzzer-sys` and see what's up.

nagisa:
Memleak is detected always, even for regular fuzzing. It's a known issue.
As for warning it is intended. Running with a single input runs test with
that input only, and not fuzzing which also sounds intended.

neosilky:
@nagisa Sure - that's okay then. This should still be okay to merge through, right?

PR #106, opened 2017-04-27
Support macOS

### LINKED ISSUE #36: Support for MacOS via libfuzzer
There's supposedly support for libfuzzer on MacOS in the latest LLVM (I think?). Opening this here as a tracking issue and reminding us we might need to change [what was added here](https://github.com/rust-fuzz/cargo-fuzz/pull/34).

kennytm:
Fix #36. 

rustc enabled ASan and TSan on macOS starting from 1.18.0-nightly (036983201 2017-04-26), so no special change is needed to enable macOS support, aside from relaxing the target check.

This PR adds the `--target` flag to let the user specify the target triple. By default, the target is set to `x86_64-apple-darwin` on macOS, and `x86_64-unknown-linux-gnu` everywhere else, to avoid  changing too much from #34.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -53,6 +53,8 @@ fn main() {
                  .help("custom corpus directory or artefact files"))
             .arg(Arg::with_name("ARGS").multiple(true).last(true)
                  .help("additional libFuzzer arguments passed to the binary"))
+            .arg(Arg::with_name("TRIPLE").long("target")
+                 .help("target triple of the fuzz target"))
         )
         .subcommand(SubCommand::with_name("add").about("Add a new fuzz target")
                     .arg(Arg::with_name("TARGET").required(true)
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -53,6 +53,8 @@ fn main() {
                  .help("custom corpus directory or artefact files"))
             .arg(Arg::with_name("ARGS").multiple(true).last(true)
                  .help("additional libFuzzer arguments passed to the binary"))
+            .arg(Arg::with_name("TRIPLE").long("target")
```
nagisa: This should be `.takes_value(true)` and default value should also be specified here and not when the value is unwrapped.

/me smacks @Manishearth for merging this.

frewsxcv:
Why not just open a PR instead of smacking people

Manishearth:
lol

PR #109, opened 2017-05-08
Implement parallel fuzzing

nagisa:
It is possible to run multiple fuzzing processes at the same time, thus
improving fuzzing speed signinficantly. One problem with the built-in
fuzzing is that it is bad at displaying information on the go.

Therefore we implement our own process mangement (using the awesome
tokio) and print results from all of them on the fly.

Also fixes the CLI for `--target`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,13 +8,22 @@
 extern crate toml;
 extern crate clap;
 extern crate term;
+extern crate tokio_core;
+extern crate tokio_process;
+extern crate tokio_io;
+extern crate futures;
 #[macro_use]
 extern crate error_chain;
 
 use clap::{App, Arg, SubCommand, ArgMatches, AppSettings};
-use std::{env, fs, path, ffi, process};
+use std::{env, fs, path, ffi, process, io};
 use std::io::Write;
 use std::io::Read;
+use std::ffi::OsStr;
+
+use futures::{Future, Stream};
+use tokio_core::reactor::Core;
+use tokio_process::CommandExt;
 
 #[macro_use]
 mod templates;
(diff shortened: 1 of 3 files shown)


nagisa:
cc @Manishearth @frewsxcv for a review.

src/main.rs
```diff
@@ -51,10 +60,21 @@ fn main() {
                  .help("name of the fuzz target"))
             .arg(Arg::with_name("CORPUS").multiple(true)
                  .help("custom corpus directory or artefact files"))
-            .arg(Arg::with_name("ARGS").multiple(true).last(true)
-                 .help("additional libFuzzer arguments passed to the binary"))
             .arg(Arg::with_name("TRIPLE").long("target")
+                 .takes_value(true)
+                 .default_value(utils::default_target())
```
frewsxcv: not a big deal, but `default_value` implies `takes_value`

https://docs.rs/clap/2.24.2/clap/struct.Arg.html#method.default_value

src/main.rs
```diff
@@ -245,8 +282,48 @@ impl FuzzProject {
             cmd.arg(self.corpus_for(&target)?);
         }
 
-        exec_cmd(&mut cmd).chain_err(|| format!("could not execute command: {:?}", cmd))?;
-        Ok(())
+        if jobs == 1 {
+            exec_cmd(&mut cmd).chain_err(|| format!("could not execute command: {:?}", cmd))?;
+            Ok(())
```
frewsxcv: could `return` here if you wanted to avoid the indentation in the `else` block, up to you

nagisa:
bors r=frewsxcv

PR #114, opened 2017-06-02
Add commands to minimize corpuses and test cases

alanhdu:
This adds two commands:
- `cargo fuzz tmin TARGET TEST_CASE` will try and minimize a crashing test case.
- `cargo fuzz cmin TARGET <CORPUS>` will minimize a given corpus while keeping coverage the same.

The names are based off of `afl-tmin` and `afl-cmin`, which come bundled with `afl-fuzz`.


alanhdu:
So the corpus minification actually presents a different interface from what `libfuzzer` provides -- `libfuzzer` lets you merge new test cases that add coverage from one directory to another, while I wrote `cmin` to just minimize a given corpus (which is accomplished with a bunch of temporary directories).

I also changed the default settings so that fuzzers build with optimization and debug asserts by default. I personally think that those defaults make more sense for fuzzing, but I'm happy to back that commit out if there are problems.

src/main.rs
```diff
@@ -65,18 +69,38 @@ fn main() {
...
         (s, _) => panic!("unimplemented subcommand {}!", s),
     }.map(|_| 0).unwrap_or_else(|err| {
         utils::report_error(&err);
         1
     }));
 }
 
+fn fuzz_wrapper(name: &str) -> App {
+    SubCommand::with_name(name)
+        .arg(Arg::with_name("no_opt").long("no-opt").short("no")
```
nagisa: Following two arguments change the current CLI API unnecessarily.

src/main.rs
```diff
@@ -137,12 +161,11 @@ impl FuzzProject {
     }
 
     fn add_target(&self, args: &ArgMatches) -> Result<()> {
-        let target: String = args.value_of_os("TARGET").expect("TARGET is required").to_os_string()
-            .into_string().map_err(|_| "TARGET must be valid unicode")?;
+        let target = args.value_of("TARGET").expect("No TARGET");
```
nagisa: This makes the error worse in case the TARGET is not valid unicode.

src/main.rs
```diff
@@ -158,29 +181,24 @@ impl FuzzProject {
...
-    /// Fuzz a given fuzz target
-    fn exec_target<'a>(&self, args: &ArgMatches<'a>) -> Result<()> {
-        let target: String = args.value_of_os("TARGET").expect("TARGET is required").to_os_string()
-            .into_string().map_err(|_| "TARGET must be valid unicode")?;
-        let release: bool = args.is_present("release");
-        let assertions: bool = args.is_present("debug_assertions");
+
+    /// Get fuzzer command
+    fn cmd(&self, args: &ArgMatches) -> Result<process::Command> {
+        let target = args.value_of("TARGET").expect("TARGET is required");
```
nagisa: Same here.

src/main.rs
```diff
@@ -220,32 +238,85 @@ impl FuzzProject {
...
+        if !args.is_present("no_opt") {
             cmd.arg("--release");
         }
         cmd.arg("--verbose")
            .arg("--bin")
            .arg(&target)
            .arg("--target")
-           // won't pass rustflags to build scripts
-           .arg(target_triple)
+           .arg(utils::default_target()) // won't pass rustflags to build scripts
```
nagisa: This seems to change sementics significantly? Possibly wrong too?

alanhdu: Kind of -- this just uses the default triple target (`x86_64-apple-darwin` or `x86_64-unknown-linux-gnu`) instead of letting the user pass a triple target in.

The thing is, there's actually no way to pass in triple target right now anyways 😸 because the corresponding flag doesn't have a `takes_value(true)` (see https://github.com/rust-fuzz/cargo-fuzz/blob/master/src/main.rs#L57)!

Seeing as no one's run into this bug in the last month, I kind of figured that no one actually needed to override the default target and decided to just remove that functionality (especially becuase you can just set the `RUSTFLAGS` environment variable if you really need to). Of course, if you think it's important that we support this, I can also just add the `takes_value(true)` and fix the underlying bug.

src/main.rs
alanhdu: So of this change was to switch what happens when you just run `cargo fuzz run TARGET` -- right now, this fuzzes an unoptimized binary with debug assertions whereas IMO we should default to fuzzing an optimized binary with debug assertions, and I'm not sure how else to do that without changing the CLI.

This also gives the user strictly more flexibility -- right now, there's no way to build an unoptimized binary without debug assertions (although I'm not sure why you'd want to 😜), whereas this change lets you do all 4 combinations of unoptimized/optimized and debug asserts / no debug asserts.

src/main.rs
nagisa: > The thing is, there's actually no way to pass in triple target right now anyways because the corresponding flag doesn't have a takes_value(true) (see https://github.com/rust-fuzz/cargo-fuzz/blob/master/src/main.rs#L57) 😸!

This is a bug that was fixed in https://github.com/rust-fuzz/cargo-fuzz/pull/109 which I simply forgot to merge before today.

alanhdu: Cool cool. I'll back that out while I'm resolving this merge conflict.

alanhdu:
@nagisa: I backed out the comment changing the CLI flags and defaults -- I'll open a separate PR for that later.

src/main.rs
```diff
@@ -234,53 +266,55 @@ impl FuzzProject {
...
+            .chain_err(|| format!("could not execute: {:?}", cmd))?;
         if !status.success() {
             return Err(format!("could not build fuzz script: {:?}", cmd).into());
         }
 
-        let mut cmd = process::Command::new("cargo");
-        cmd.arg("run")
-           .args(&cargo_args);
-        for &(ref k, ref v) in &envs { cmd.env(k, v); } // Command::envs still unstable
+        let mut cmd = self.cmd(args)?;
```
alanhdu: I'm not super happy that we reparse all the CLI flags again here, but I think the DRY code savings outweigh the (hopefully) small waste.

alanhdu:
@nagisa: Do you mind taking another look at this?

src/main.rs
```diff
@@ -102,6 +88,25 @@ Some useful options (to be used as `cargo fuzz run fuzz_target -- <options>`) in
...
+             .about("Corpus minifier")
+             .arg(Arg::with_name("CORPUS")
+                  .help("directory with corpus to minify"))
+        )
+        .subcommand(fuzz_subcommand("tmin")
+             .about("Test case minifier")
+             .arg(Arg::with_name("runs").long("runs")
+                  .help("number of attempts to minimize we should make")
+                  .takes_value(true)
+                  .default_value("255")
```
nagisa: Is 255 enough to minimise anything non-trivial? 255 cases is seems like something that fuzzer could easily do in half a second so raising this value could be interesting.

src/main.rs
nagisa: Yeah. There’s very little CLI flag reparsing going on, so it doesn’t matter much. Most of the work is already done by the time we get ArgMatches.

src/main.rs
alanhdu: I've actually had pretty good success with only 100 cases when I fuzzed `toml`, so I think 200 or 300 should work fine, but I have no problems bumping this up!

nagisa: I guess it is fine then.

nagisa:
bors r+

Nice job!

PR #118, opened 2017-06-10
Update the arguments passed to llvm

deweerdt:
libFuzzer expects (2ac50c9432f4856b505e3f9cb7f4e7161f5e182c 2017-06-07) more instrumentation in the binaries to be fuzzed

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -221,7 +221,8 @@ impl FuzzProject {
         let other_flags = env::var("RUSTFLAGS").unwrap_or_default();
         let mut rustflags: String = format!(
             "-Cpasses=sancov \
-             -Cllvm-args=-sanitizer-coverage-level=3 \
+             -Cllvm-args=-sanitizer-coverage-level=4 \
+             -Cllvm-args=-sanitizer-coverage-trace-pc \
              -Zsanitizer={sanitizer} \
              -Cpanic=abort",
             sanitizer = sanitizer,


frewsxcv:
The [errors Travis found](https://travis-ci.org/rust-fuzz/cargo-fuzz/jobs/241402282) seem relevant to these changes. Do we need to bump libfuzzer in libfuzzer-sys?

nagisa:
Yeah I guess we could try, but this needs for both projects to be updated in tandem. This is probably the first good reason I’ve seen to publish libfuzzer-sys into crates.io.

nagisa:
I will think about it tomorrow.

The PR itself looks good. Will merge… eventually.

nagisa:
Still on my radar.

deweerdt:
I just stumbled on [this](http://llvm.org/docs/LibFuzzer.html):  "Very recent versions of Clang (> April 20 2017) include libFuzzer, and no installation is necessary. In order to fuzz your binary, use the -fsanitize=fuzzer flag during the compilation"

That sounds like an improvement in usability, maybe it's worth to wait a bit for things to setting in libFuzzer land.

nagisa:
bors try

nagisa:
We seem to have updated our flags to a larger set independently after updating to the newer libfuzzer and LLVM.

Thanks for the PR!

PR #123, opened 2017-06-19
Fix typo (artefact -> artifact)

danburkert:


8573:
I think that's unlikely to be a typo, and more likely an intentional,
or unthinking, use of the British spelling of the word.

frewsxcv:
Ah, thanks for pointing that out @8573, I didn't know there was an alternate spelling. Looks like [libfuzzer](http://llvm.org/docs/LibFuzzer.html) uses the spelling with the 'i', so I'm going to keep it as is

Issue #100, opened 2017-04-04
Shadow memory range interleaves with an existing memory mapping. ASan cannot proceed correctly. ABORTING


chifflier:
Hi,

cargo-fuzz is not working anymore here, but I cannot find the source of the problem. The error is triggered when starting the fuzzer:
```
$ cargo fuzz run fuzzer_script_1
...
     Running `fuzz/target/x86_64-unknown-linux-gnu/debug/fuzzer_script_1 -artifact_prefix=/home/pollux/CODE/RUST/der-parser/fuzz/artifacts/fuzzer_script_1/ /home/pollux/CODE/RUST/der-parser/fuzz/corpus/fuzzer_script_1`
==6747==Shadow memory range interleaves with an existing memory mapping. ASan cannot proceed correctly. ABORTING.
==6747==ASan shadow was supposed to be located in the [0x00007fff7000-0x10007fff7fff] range.
==6747==Process memory map follows:
	0x0005eba8d000-0x0005ebc94000	/home/pollux/CODE/RUST/der-parser/fuzz/target/x86_64-unknown-linux-gnu/debug/fuzzer_script_1
	0x0005ebe93000-0x0005ebe98000	/home/pollux/CODE/RUST/der-parser/fuzz/target/x86_64-unknown-linux-gnu/debug/fuzzer_script_1
	0x0005ebe98000-0x0005ebe9f000	/home/pollux/CODE/RUST/der-parser/fuzz/target/x86_64-unknown-linux-gnu/debug/fuzzer_script_1
	0x0005ebe9f000-0x0005edd2a000	
	0x0005edd2a000-0x0005f18ed000	
	0x0005f18ed000-0x0005f18ee000	[heap]
	0x03a358fd1000-0x03a359323000	
	0x03a359323000-0x03a359426000	/lib/x86_64-linux-gnu/libm-2.24.so
	0x03a359426000-0x03a359625000	/lib/x86_64-linux-gnu/libm-2.24.so
	0x03a359625000-0x03a359626000	/lib/x86_64-linux-gnu/libm-2.24.so
	0x03a359626000-0x03a359627000	/lib/x86_64-linux-gnu/libm-2.24.so
	0x03a359627000-0x03a3597bc000	/lib/x86_64-linux-gnu/libc-2.24.so
	0x03a3597bc000-0x03a3599bb000	/lib/x86_64-linux-gnu/libc-2.24.so
[...]

PR #126, opened 2017-07-30
workaround for issue #100

bennofs:
Using link-args, we force the load address for the executable to be higher than
the shadow memory area required by address sanitizer.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -311,6 +311,7 @@ impl FuzzProject {
              -Cpasses=sancov \
              -Cllvm-args=-sanitizer-coverage-level=3 \
              -Zsanitizer={sanitizer} \
+             -Clink-args=-Wl,-Ttext-segment=0x200000000000 \
              -Cpanic=abort",
             sanitizer = sanitizer,
         );


frewsxcv:
@manishearth @nagisa Are either of you familiar with what's going on here? I'm not the best reviewer for this

Manishearth:
not entirely. Makes some sense looking at https://github.com/google/sanitizers/issues/837 but I'm not entirely sure this is the right fix.

bennofs:
What this commit does is the following. Usually, the memory map of a rust exe at runtime would looks like this:

```
[0x00007fff7000-0x10007fff7fff]  shadow memory region for asan
ELF_ET_DYN_BASE + random offset: text segment (contains the executable code)
directly afterwards: data segment (contains global variables etc)
... lots of free address space ...
dynamic libraries (mmap region)
```

The problem now is that `ELF_ET_DYN_BASE` is so low that it is in most cases (unless you get lucky with the ASLR offset) is lower than the top end of the shadow memory region, so it collides. The "fix" forces GCC to place the text segment at a fixed, higher address so that it cannot collide with the shadow memory region. With the fix, the memory map would look like this:

```
[0x00007fff7000-0x10007fff7fff]  shadow memory region for asan
0x200000000000 (not random): text segment (contains the executable code)
directly afterwards: data segment (contains global variables etc)
... lots of free address space ...
dynamic libraries (mmap region)
```

Note that this means that the code is always loaded at the same address (this makes exploiting buffer overflows etc easier, but it should not be a problem since fuzzing is not run against user data?), but it guarantees that it won't collid with the shadow memory space.

Possible questions:

* I didn't find any option to control the data segment. It seems (?) that the data segment is always loaded directly following the text segment, so we are safe if the text segment is placed high enough
* Ideally, we would only apply this linker option to the final executable, as I am not sure how it interacts with a `dylib` crates (they of course cannot be placed at a fixed location). But I couldn't get `dylib` crates to work with ASan even without this fix, so I cannot test this.


Other possible solutions:

* Another solution would be to disable position independent executables, and hope that the default load address of the executable is high enough (which appears to be the case right now). It seems that you can do that with `-C relocation-model=dynamic-no-pic`. I did not go for this solution though as I think it is less reliable than simply specifying the load address thus ensuring that it can never collide.

nagisa:
I’m generally against such workarounds for several reasons:

1. It is super easy to forget to remove a workaround when it is not necessary anymore;
2. I generally consider it to be a great idea to fix issues like these upstream – either in the kernel, if it is a kernel bug (which seems unlikely), or the sanitizer (why is it assuming *anything* about any addresses??);
3. It is non-portable for the eventual future when all this stuff works on platforms other than Linux. It also assumes a particular linker which supports this particular operation/flag.

OTOH this comes with benefits as well, such as disablement of ASLR which results in more reproducible runs, so I’m only slightly against.

bennofs:
@nagisa I'm not sure that this can be fixed in the sanitizer / kernel. ASan needs a bit of memory for book keeping, and it appears that hardcoding this location is necessary for performance (upstream mentioned that they did not want to make it configurable because performance would suffer too much). So the only "upstream" that could fix this would be rustc.

frewsxcv:
Anyone opposed to closing this?

bennofs:
Has the bug been fixed? It's been a long time, I have not kept up with the latest developments in these areas. 

However, I guess this particular PR can be closed anyhow, since that's not the approach anybody seems to want to take.

PR #134, opened 2017-11-10
Add badges

sanmai-NL:
This helps promotion/adoption of `cargo-fuzz`. Also, I wanted to ask a few quick questions but couldn't discover a chat room very quickly (e.g., Google search, `README`). I'm sure you are all on IRC ... But I hope you agree the Gitter channel won't hurt: non-maintainers can answer each other's questions that way as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,15 @@
 # Cargo-fuzz
 
+
+[![MIT license](https://img.shields.io/badge/license-MIT-blue.svg)](./LICENSE-MIT)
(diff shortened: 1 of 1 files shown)


frewsxcv:
Thanks for the pull request! I don't think the badges are super necessary, in particular the API docs and gitter links (which are both broken) and the license is already mentioned in the README. But I don't feeel particularly strongly about this so someone else in this GitHub org can decide

sanmai-NL:
I find it wasteful how this contributions was handled and the motivation is unclear (‘don’t think they are super necessary’). But I’m sorry to say, not going to spend more energy on it.

PR #142, opened 2018-02-04
Re-enable -sanitizer-coverage-trace-pc-guard for macOS.

frewsxcv:
The rustc bug can be avoided by specifying codegen units to 1.


frewsxcv:
hah just kidding i was mixing up things

PR #147, opened 2018-03-27
Enable '-sanitizer-coverage-trace-pc-guard' on macOS.

frewsxcv:


frewsxcv:
this is now fixed on macOS upstrream, so we don't need this conditional anymorer

PR #150, opened 2018-04-23
Commit Cargo.lock

Ekleog:
As an executable, I think `cargo-fuzz` should include its `Cargo.lock` in its repository. This will also allow for easy packaging in eg. nixpkgs, which is the reason why I noticed the `Cargo.lock` file was absent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
-Cargo.lock
 target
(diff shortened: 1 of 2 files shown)


nagisa:
Seems good to me.

bors r+

PR #184, opened 2019-10-20
Use ? instead of try!

alex:


alex:
@nagisa can you redo this one, it needed to land after the sancov flag fixes

PR #204, opened 2019-12-19
Build subcommand + running fuzz targets on specific inputs

fitzgen:
r? @Manishearth 

* Add tests for `cargo fuzz run <file>...`
  
  So this didn't used to work, as there was what I think was a bug where the
  default corpus was unconditionally added to the underlying command. That got
  fixed in an earlier refactoring, so this just makes sure we never regress the
  intended behavior with some tests.

* Introduce the `cargo fuzz build` subcommand

* Pull `target` out of `BuildOptions`
  
  Building in general doesn't require that you only build a single target.

* Move `BuildOptions` to the `options` submodule
  
  Just code motion.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,4 +15,5 @@ jobs:
       run: |
         rustup install nightly
         rustup default nightly
+        export RUST_BACKTRACE=1
         cargo test --verbose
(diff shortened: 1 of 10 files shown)


fitzgen:
Thanks!

PR #215, opened 2020-02-01
Force 1 CGU when release mode is enabled

alexcrichton:
We've locally started setting up OSS fuzz and Rust recently, but we're
unfortunately getting a lot of timeouts where tests are taking much
longer than expected. After some investigation one cause of the issue
seems to be that although we're compiling the fuzzers in release mode
they're still performing quite badly at runtime.

One cause of this appears to be that functions are not being inlined
across CGUs within a crate, and digging even further seems to show that
functions are tagged with `notEligibleToImport` which presumably means
that ThinLTO cannot import them across CGUs.

Poking around in LLVM it looks like the sancov passes add mutations of a
private `__sancov_gen_` global in all functions in a CGU, and then that
global itself is listed as not importable. Any function referencing
something not importable is then also considered not importable, hence
nothing in any CGU is importable.

With this change, however, the performance of optimized-Rust being
sanitized is unfortunately still ~100x slower than non-sanitized Rust. I
don't think the overhead is supposed to be that high so I'm still
looking into other causes of slowness.

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -174,6 +174,18 @@ impl FuzzProject {
             rustflags.push_str(" -Cdebug-assertions");
         }
 
+        // If release mode is enabled then we force 1 CGU to be used in rustc.
+        // This will result in slower compilations but it looks like the sancov
+        // passes otherwise add `notEligibleToImport` annotations to functions
+        // in LLVM IR, meaning that *nothing* can get imported with ThinLTO.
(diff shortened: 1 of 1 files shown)


alexcrichton:
cc @fitzgen, this is related to some of the timeout bugs that we were looking at for wasmtime

nagisa:
Definitely makes sense to do this.

bors r+

Issue #220, opened 2020-03-17
Add a `debug-fmt` subcommand to print the std::fmt::Debug output of a libfuzzer input


fitzgen:
Would be used like:

```
$ cargo fuzz debug-fmt <target> <input>
<output of std::fmt::Debug printed to stdout>
```

For example, a fuzz target that takes an `Rgb` struct, turns it to HSL, and back would output something like:

```
$ cargo fuzz debug-fmt rgb_to_hsl_to_rgb corpus/rgb_to_hsl_to_rgb/abcdefghijklmnopqrstuvwxyz
Rgb {
    r: 12,
    g: 33,
    b: 127,
}
```

[This would use the existing `RUST_LIBFUZZER_DEBUG_PATH=...` environment variable that we use for when a fuzz run fails.](https://github.com/rust-fuzz/cargo-fuzz/blob/2243de096b15b79b719ce7489f014d7d8ce197ee/src/project.rs#L295-L334)

fitzgen:
Fixed in #221

PR #221, opened 2020-03-21
Adds support for `fmt` subcomand for test case debug formatting

ex0dus-0x:
Adds support for https://github.com/rust-fuzz/cargo-fuzz/issues/220.

Subcommand routine is pretty rudimentary, where it basically checks to see if the test case exists, and then calls `run_fuzz_target_debug_formatter` to properly output the test case (structured input if `arbitrary` is implemented).

Let me know if anything else should be implemented!

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -336,19 +336,26 @@ impl FuzzProject {
     /// Prints the debug output of an input test case
     pub fn debug_fmt_input(&self, debugfmt: &options::Fmt) -> Result<()> {
         if !debugfmt.input.exists() {
-            bail!("Input test case does not exist.");
+            bail!(
+                "Input test case does not exist: {}",
+                debugfmt.input.display()
+            );
         }
 
-        if let Ok(debug) =
-            self.run_fuzz_target_debug_formatter(&debugfmt.build, &debugfmt.target, &debugfmt.input)
-        {
-            eprintln!("Output of `std::fmt::Debug`:\n");
-            for l in debug.lines() {
-                eprintln!("{}", l);
-            }
-        } else {
-            bail!("Unable to return debug-formatted output for input test case.")
+        let debug = self
+            .run_fuzz_target_debug_formatter(&debugfmt.build, &debugfmt.target, &debugfmt.input)
+            .with_context(|| {
+                format!(
+                    "failed to run `cargo fuzz fmt` on input: {}",
+                    debugfmt.input.display()
+                )
+            })?;
+
+        eprintln!("\nOutput of `std::fmt::Debug`:\n");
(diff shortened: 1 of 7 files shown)


fitzgen:
[Can you run `cargo fmt` and update this PR? Thanks!](https://github.com/rust-fuzz/cargo-fuzz/pull/221/checks?check_run_id=523425810#step:3:19)

Manishearth:
`fmt` works for me

src/project.rs
```diff
@@ -333,6 +333,25 @@ impl FuzzProject {
         Ok(debug)
     }
 
+    /// Prints the debug output of an input test case
+    pub fn debug_fmt_input(&self, debugfmt: &options::Fmt) -> Result<()> {
+        if !debugfmt.input.exists() {
+            bail!("Input test case does not exist.");
```
fitzgen:
```suggestion
            bail!("Input test case does not exist: {}", debugfmt.input.display());
```

src/project.rs
```diff
@@ -333,6 +333,25 @@ impl FuzzProject {
...
+
+        if let Ok(debug) =
+            self.run_fuzz_target_debug_formatter(&debugfmt.build, &debugfmt.target, &debugfmt.input)
+        {
+            eprintln!("Output of `std::fmt::Debug`:\n");
+            for l in debug.lines() {
+                eprintln!("{}", l);
+            }
+        } else {
+            bail!("Unable to return debug-formatted output for input test case.")
```
fitzgen: Instead of doing `if let Ok(..) = { .. } else { .. }` and creating a new error in the not-`Ok` case, let's use [the `with_context` method from `anyhow`](https://docs.rs/anyhow/1.0.27/anyhow/trait.Context.html#tymethod.with_context) to add extra info to the error returned from `run_fuzz_target_debug_formatter`:

```rust
let debug = self.run_fuzz_target_debug_formatter(&debugfmt.build, &debugfmt.target, &debugfmt.input)
    .with_context(|| format!("failed to run `cargo fuzz fmt` on input:", debugfmt.input.display()))?;
```

This should be both more concise and provide more information when errors occur.

tests/tests/main.rs
```diff
@@ -408,6 +408,62 @@ fn run_alt_corpus() {
...
+
+                #[derive(Debug)]
+                pub struct Rgb {
+                    r: u8,
+                    g: u8,
+                    b: u8,
+                }
+
+                impl Arbitrary for Rgb
+                {
```
fitzgen: Nitpick: brace on the same line as `impl Arbitrary for Rgb`

Issue #216, opened 2020-02-01
Sanitizer coverage options can slow down binaries up to 50x


alexcrichton:
Recently we've added the `wasmtime` project to OSS fuzz and we've gotten a lot of bugs (yay!). Some of the bugs are related to timeouts and when investigating locally I was surprised that running un-fuzzed code was orders of magnitude faster than the fuzzed code so it seemed that simply fuzzing caused code to get so slow it would time out (but we saw reasonable performance on the same tests outside of fuzzing).

One big chunk of the slowdown is in https://github.com/rust-fuzz/cargo-fuzz/pull/215, but it looks like the various sanitizer coverage options are also a source of quite a large slowdown.

I ran some tests locally and was executing a command like so:

```
$ cargo +nightly run --release --target x86_64-unknown-linux-gnu --bin differential ./the-test-file
```

Basically I was curious to just run the fuzzer binary on one particular test case, timing how long it took. I varied `RUSTFLAGS` according to below and got the following timings. Note that each of these has `-Ccodegen-units=1` to account for https://github.com/rust-fuzz/cargo-fuzz/pull/215.

| `RUSTFLAGS` | test time |
|---------------|-------------|
| (empty) | 89ms |
| sancov level 1 | 385ms  |
| sancov level 4 | 428ms |
| sancov level 4 + `-sanitizer-coverage-pc-table` | 353ms |
| sancov level 4 + `-sanitizer-coverage-prune-blocks=0` | 506ms |
| sancov level 4 + `-sanitizer-coverage-trace-geps` | 1708ms |
| sancov level 4 + `-sanitizer-coverage-trace-compares` | 2327ms |
| sancov level 4 + all options | 4113ms |

Note there that "sancov level 1" corresponds to `-Cpasses=sancov -Cllvm-args=-sanitizer-coverage-level=1 -Cllvm-args=-sanitizer-coverage-inline-8bit-counters`  and "sancov level 4" correspond to `level=4` there. The default in `cargo fuzz` right now is sancov level 4. Also note that the last row above, "sancov level 4 + all options", is the default for `cargo fuzz` today.

What I largely want to highlight here is that cranking up the sancov level and options causes the slowdown of a binary to get quite large to the point that a timeout on a fuzzer may not be too relevant if it's executing up to 50x slower than it would otherwise. This leads me to a few questions:

* What are these sancov options and why are they passed by default? Do they improve libfuzzer's discovery of test cases?
[...]

PR #229, opened 2020-06-09
Remove some LLVM options which excessively slow down code

alexcrichton:
This commit is a bit of a stab towards #216. I discussed this a bit more
on google/oss-fuzz#3944 and this is intended to sync Rust's `cargo fuzz`
with the default C++ fuzzing there. Apparently the `prune-blocks` and
`trace-geps` options aren't used by libFuzzer and are either
experimental or disable too many optimizations.


fitzgen:
+1 from me too -- gonna merge this as three people signing off seems like plenty :)

I can cut a new release as well.

fitzgen:
Thanks for investigating and following up, Alex!

fitzgen:
https://crates.io/crates/cargo-fuzz/0.7.5

Issue #228, opened 2020-06-05
Recommended reproduction command ignores arguments


HeroicKatora:
When running the fuzzer with specific sanitizer, time or memory settings and an issue is found then the output for reproduction is not complete. In particular the recommended command line for reproducing the issue misses any of these parameters as well as the used `cargo` version. Abbreviated example below.

```sh
$ cargo +nightly fuzz run some_fuzzer -sanitizer=memory
# ...
SUMMARY: MemorySanitizer: use-of-uninitialized-value (...)
Exiting
# ...
Reproduce with:

    cargo fuzz run some_fuzzer fuzz/artifacts/parser/crash-hash
```

When it should instead print:

```sh
Reproduce with:

    cargo +nightly fuzz run some_fuzzer -sanitizer=memory fuzz/artifacts/parser/crash-hash
```

frewsxcv:
Thanks for reporting! Here's the code that needs to change:

https://github.com/rust-fuzz/cargo-fuzz/blob/2fb3bd85c0dd57ea7f3d5c29174c26a5550d0aa8/src/project.rs#L426-L436

We need to incorporate [the arguments](https://github.com/rust-fuzz/cargo-fuzz/blob/2fb3bd85c0dd57ea7f3d5c29174c26a5550d0aa8/src/options/run.rs#L30-L32) into the output.

ian-fox:
> Thanks for reporting! Here's the code that needs to change:
> 
> https://github.com/rust-fuzz/cargo-fuzz/blob/2fb3bd85c0dd57ea7f3d5c29174c26a5550d0aa8/src/project.rs#L426-L436
> 
> We need to incorporate [the arguments](https://github.com/rust-fuzz/cargo-fuzz/blob/2fb3bd85c0dd57ea7f3d5c29174c26a5550d0aa8/src/options/run.rs#L30-L32) into the output.

I think the things we actually need here are from the `BuildOptions`. `args` holds things like `-runs` but not which sanitizers were used.

Unsure so far how to get the `+nightly` bit, I'll take a closer look at that tomorrow.

nagisa:
`+nightly` comes from `rustup` wrapper, so I suspect its going to be out-of-scope for `cargo-fuzz`. But we can probably add a note to that effect? Like "note: if you're using rustup, you may need to specify `+nightly` to the cargo command".

ian-fox:
That's probably the easiest way to do it. 

[...]

PR #231, opened 2020-06-20
Faster defaults

### LINKED ISSUE #230: Fuzzing with default settings is much slower than it needs to be
`cargo fuzz run` currently defaults to debug mode with Address Sanitizer. Debug mode in Rust is needlessly slow, easily 20x slower than release mode, and for fuzzing performance is critical. It would be a much better idea to default to release mode, but with debug assertions and overflow checks enabled.

The only reason to invoke actual debug mode is when reproducing the testcase and actually trying to debug it. So it'd be nice to add a flag `--debug` that forces debug mode and print it in the crash output, but this should be a non-breaking change in all other respects.

Shnatsel:
* Default to compiling in release mode with overflow checks and debug assertions enabled
    * Speeds up fuzzing ~10x
* ~~Add command line option for controlling overflow checks~~
* Add command line option for forcing debug mode without optimizations

Resolves #230

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -40,7 +40,11 @@ fuzz/fuzz_targets/, i.e. the name picked when running `cargo fuzz add`.
 This will run the script inside the fuzz target with varying inputs until it
 finds a crash, at which point it will save the crash input to the artifact
 directory, print some output, and exit. Unless you configure it otherwise (see
-libFuzzer options below), this will run indefinitely.";
+libFuzzer options below), this will run indefinitely.
+
+By default fuzz targets are built with optimizations equivalent to
+`cargo build --release`, but with debug assertions and overflow checks enabled.
+Address Sanitizer is also enabled by default.";
 
 const RUN_AFTER_HELP: &'static str = "\
 
@@ -66,6 +70,17 @@ include:
 
   * `-dict=<file>`: Use a keyword dictionary from specified file. See
     http://llvm.org/docs/LibFuzzer.html#dictionaries\
+";
+
+const BUILD_BEFORE_HELP: &'static str = "\
+By default fuzz targets are built with optimizations equivalent to
+`cargo build --release`, but with debug assertions and overflow checks enabled.
+Address Sanitizer is also enabled by default.";
+
+const BUILD_AFTER_HELP: &'static str = "\
+Sanitizers perform checks necessary for detecting bugs in unsafe code
+at the cost of some performance. For more information on sanitizers see
+https://doc.rust-lang.org/unstable-book/compiler-flags/sanitizer.html\
 ";
 
 /// A trait for running our various commands.
@@ -94,6 +109,11 @@ enum Command {
     /// Add a new fuzz target
     Add(options::Add),
 
+    #[structopt(
+        template(LONG_ABOUT_TEMPLATE),
+        before_help(BUILD_BEFORE_HELP),
+        after_help(BUILD_AFTER_HELP)
+    )]
     /// Build fuzz targets
     Build(options::Build),
 
--- a/src/options.rs
+++ b/src/options.rs
@@ -66,7 +66,7 @@ pub struct BuildOptions {
     pub release: bool,
 
     #[structopt(short = "a", long = "debug-assertions")]
-    /// Build artifacts with debug assertions enabled (default if not -O)
+    /// Build artifacts with debug assertions and overflow checks enabled (default if not -O)
     pub debug_assertions: bool,
 
     /// Build target with verbose output from `cargo build`
--- a/src/options.rs
+++ b/src/options.rs
@@ -69,10 +69,6 @@ pub struct BuildOptions {
     /// Build artifacts with debug assertions enabled (default if not -O)
     pub debug_assertions: bool,
 
-    #[structopt(short = "o", long = "overflow-checks")]
-    /// Build artifacts with overflow checks enabled (default if not -O)
-    pub overflow_checks: bool,
-
     /// Build target with verbose output from `cargo build`
     #[structopt(short = "v", long = "verbose")]
     pub verbose: bool,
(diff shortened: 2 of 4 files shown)


src/options.rs
```diff
@@ -57,18 +57,26 @@ impl FromStr for Sanitizer {
 
 #[derive(Clone, Debug, StructOpt)]
 pub struct BuildOptions {
-    #[structopt(short = "O", long = "release")]
+    #[structopt(short = "D", long = "debug", conflicts_with = "release")]
+    /// Build artifacts in debug mode, without optimizations
+    pub debug: bool,
```
fitzgen, line 62: Let's name this `--dev` to match the cargo profile name, and to avoid confusion like "how is this different from `--debug-assertions`?"

src/options.rs
```diff
@@ -57,18 +57,26 @@ impl FromStr for Sanitizer {
...
-    /// Build target with verbose output from `cargo build`
-    #[structopt(short = "v", long = "verbose")]
-    pub verbose: bool,
-
     #[structopt(short = "a", long = "debug-assertions")]
     /// Build artifacts with debug assertions enabled (default if not -O)
     pub debug_assertions: bool,
 
+    #[structopt(short = "o", long = "overflow-checks")]
+    /// Build artifacts with overflow checks enabled (default if not -O)
```
fitzgen, line 73: Regarding "default if not -O": right now, passing `-Oa` will enable overflow checks, is that not the case anymore after this PR? Many projects use `-Oa` and I'd be wary of making it lose the overflow checks. Do `debug-assertions` imply `overflow-checks` or are they disconnected now?

src/project.rs
```diff
@@ -507,7 +507,7 @@ impl FuzzProject {
             }
 
             eprintln!(
-                "Reproduce with:\n\n\tcargo fuzz run {target} {artifact}\n",
+                "Reproduce with:\n\n\tcargo fuzz run --debug {target} {artifact}\n",
```
fitzgen, line 510: Let's leave these without any flags for simplicity, and handle adding flags for them in #228

fitzgen:
Tests are failing because they [look for bins at the `target/debug` path](https://github.com/rust-fuzz/cargo-fuzz/blob/87c36a1deee8ca9b550fe034623a7445d7e8c770/tests/tests/main.rs#L622) in various places, so they need to update to `target/release`.

Also, please add a test for the `--dev` (ne `--debug`) flag.

src/options.rs
Shnatsel: > right now, passing -Oa will enable overflow checks, is that not the case anymore after this PR?

That is still the case. The way these flags work is unchanged. Only the default state is altered, which is now equivalent to `-Ooa`

> Do debug-assertions imply overflow-checks or are they disconnected now?

I believe debug-assertions never implied overflow-checks, and enabling overflow-checks in release mode was not possible prior to this PR.

src/project.rs
Shnatsel: Done

src/options.rs
Shnatsel: Done

Shnatsel:
I've fixed everything as requested. Will add a test for `--dev` shortly.

Do you have any ideas on how to clearly document that the default is `--release --debug-assertions --overflow-checks`? If I invoke `--help` it's not clear to me what the build mode will be if no parameters are passed.

Also, there are two changes that might be unexpected for end users:

 - This changes the coverage numbers. Now that a great deal of branches can be optimized out, the reported coverage is much lower (~2000 instead of ~7500 edges for `png` crate).
 - This changes the default location of the binary. If something was invoking the generated binary directly instead of via `cargo fuzz run`, that workflow will break.

Should we document this somewhere?

src/options.rs
Shnatsel: Oh wait, nevermind, I was confused. I'll have to test whether `-a` actually enables overflow checks or not.

Shnatsel: Okay, I've tested it and turns out `--debug-assertions` already enables overflow checks, so there's no real need to have a separate flag for it. I'll drop the extra flag.

Shnatsel:
Should be good to go now. Please take another look.

fitzgen:
> Also, there are two changes that might be unexpected for end users:

I'm personally not too concerned with either of these, but we can bump to 0.8.x to be safe. @Manishearth any opinion here?

> Do you have any ideas on how to clearly document that the default is `--release --debug-assertions --overflow-checks`? If I invoke `--help` it's not clear to me what the build mode will be if no parameters are passed.

Perhaps in the before-help blurb for `cargo fuzz run`? https://github.com/rust-fuzz/cargo-fuzz/blob/87c36a1deee8ca9b550fe034623a7445d7e8c770/src/main.rs#L36-L43

And I think we would also want a custom template for `cargo fuzz build` like how `cargo fuzz run` has: https://github.com/rust-fuzz/cargo-fuzz/blob/87c36a1deee8ca9b550fe034623a7445d7e8c770/src/main.rs#L107

This way we could include the same blurb for both subcommands's help messages.

nagisa:
> I'm personally not too concerned with either of these, but we can bump to 0.8.x to be safe.

There's not really a good reason to _not_ bump a "breaking" version number for binaries once there are larger changes to the interface. I vote bump.

Shnatsel:
I've added a blurb about default build flags to the help text of both `cargo fuzz build` and `cargo fuzz run`. I've also added a blurb about sanitizers to `cargo fuzz build` after-help text, mostly so that I wouldn't have to create yet another help template.

Manishearth:
@fitzgen I don't consider these to require a new major version

src/main.rs
```diff
@@ -68,6 +72,17 @@ include:
     http://llvm.org/docs/LibFuzzer.html#dictionaries\
 ";
 
+const BUILD_BEFORE_HELP: &'static str = "\
+By default fuzz targets are built with optimizations equivalent to
+`cargo build --release`, but with debug assertions and overflow checks enabled.
```
frewsxcv, line 77: > but with debug assertions and overflow checks enabled

which code in this pr is responsible for enabling debug assertions with release mode by default?

PR #234, opened 2020-07-02
Include build options with recommended reproduction command

### LINKED ISSUE #228: Recommended reproduction command ignores arguments

ian-fox:
Manually implementing `Display` for `BuildOptions` admittedly doesn't seem like the cleanest way of accomplishing this at first, but I think this is probably more maintainable than trying to write some sort of macro that interfaces with `StructOpt` to see what things are defaults and what the flags are.

Downsides are that if `BuildOptions` are ever changed, the change will have to be added to the `BuildOptions.fmt(Formatter)` function. If that does ever happen the unit test should catch it though.

The bit about the toolchain is contained to one commit at the end so it's easy to remove, should the decision be that we don't want it there.

Closes #228

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -55,7 +55,7 @@ impl FromStr for Sanitizer {
     }
 }
 
-#[derive(Clone, Debug, StructOpt)]
+#[derive(Clone, Debug, StructOpt, PartialEq)]
 pub struct BuildOptions {
     #[structopt(short = "D", long = "dev", conflicts_with = "release")]
     /// Build artifacts in development mode, without optimizations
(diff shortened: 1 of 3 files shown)


ian-fox:
Currently trying to figure out the best way to do that. 

Unfortunately it looks like we can't actually call the functions from `rustup` directly to get this since it isn't on `crates.io` (we could import it as a git dependency, but then _we_ wouldn't be able to publish to `crates.io`). 

So we'd have to either call it in an external process (and parse the result) or read `$RUSTUP_HOME/.rustup/settings.toml` (which does have the info we want, but is also explicitly not part of the public interface of rustup).

With the separate process method it's a little clunky, but I think doable, and I think I trust that more than reading the `settings` file directly. I'm actually kind of tempted to pull it out into its own crate, both because it seems like something I may do again in the future and to get some experience publishing crates.

ian-fox:
Made https://github.com/ian-fox/cargo-toolchain, more for my own amusement than anything else. We could pull that in or just duplicate the code if we don't want another dependency, there was less code than I expected.

ian-fox:
Parsing the result wasn't actually so bad, I'll throw it in a util module or something tonight.

Manishearth:
I am also okay with omitting the +nightly unconditionally, people can figure that part out

fitzgen:
> I am also okay with omitting the +nightly unconditionally, people can figure that part out

Yeah, this seems fine to me since `cargo fuzz` only works with nightly anyways.

ian-fox:
Sounds good to me, I kept the toolchain stuff in one commit so it'll be easy to remove that part. 

Mostly for my own interest I've been working on https://github.com/ian-fox/cargo-toolchain (though it's not yet quite as polished as I think I'd like it to be before anything should actually depend on it) but if people decide it's a problem in the future then pulling that in should make things pretty easy (or just copying the functions from it, they ended up being way smaller than I expected)

PR #236, opened 2020-08-14
Support passing --target-dir to cargo build.

tmroeder:
This is needed for projects that need --target-dir to be set to a
non-empty value to build properly.


fitzgen (approved):
Thanks! Sorry for the delay.

FWIW, you can also set the `CARGO_TARGET_DIR` env var.

Issue #245, opened 2021-01-28
Integrate source-based code coverage into a `cargo fuzz coverage` subcommand


fitzgen:
https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html

I'm imagining this would run the whole corpus and then show you the coverage, or run fuzzing for N amount of time and then show you the coverage.

Related: #162

PR #248, opened 2021-02-16
Generating raw code-based coverage data

amaurremi:
This addresses #245. ~~It adds a `--coverage` option that generates a `.profraw` data file with source-based coverage information~~ It adds a `coverage` command that can be used to generate [source-based coverage information](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/source-based-code-coverage.html) on program inputs generated by the fuzzer. ~~However, the user still needs to manually merge/index the generated raw files into a `.profdata` file~~ It produces a `.profdata` file which can be used for detailed reports and visualization. This is described in the README.

Thanks to @abizjak for helping debug a linking issue due to incompatible LLVM coverage-instrumentation flags.


README.md
```diff
@@ -61,6 +61,51 @@ You can also always find the full command-line options that are available with
...
+## Generating code coverage information
+
+Use the `--coverage` option to generate precise
+[source-based code coverage](https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html)
+information:
+```
+$ cargo fuzz run --coverage <coverage data output file> <target>
+```
+This compiles your project using the `-Zinstrument-coverage` Rust compiler flag and generates coverage data in the
+specified file. This file can be used to generate coverage reports and visualize code-coverage information.
```
nagisa, line 73: This appears to be somewhat inaccurate. At least as per the test added below the coverage is saved to `coverage_output.prof` despite `--coverage coverage_output` being specified as the argument.

Seems to me that this detail is important enough that its worth finding a way to describe this behaviour precisely.

README.md
```diff
@@ -61,6 +61,51 @@ You can also always find the full command-line options that are available with
 $ cargo fuzz --help
 ```
 
+## Generating code coverage information
```
nagisa, line 64: This documentation is probably better suited for [the book](https://github.com/rust-fuzz/book)?

README.md
```diff
@@ -61,6 +61,51 @@ You can also always find the full command-line options that are available with
...
+[source-based code coverage](https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html)
+information:
+```
+$ cargo fuzz run --coverage <coverage data output file> <target>
+```
+This compiles your project using the `-Zinstrument-coverage` Rust compiler flag and generates coverage data in the
+specified file. This file can be used to generate coverage reports and visualize code-coverage information.
+If you run the fuzzer multiple times, you can specify different coverage-output file names and subsequently merge them
+into one data file.
+Read more in the [Unstable book](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/source-based-code-coverage.html#installing-llvm-coverage-tools).
```
nagisa, line 76: This links to the section about installing the tools, rather than merging the coverage files (which is the next section).

README.md
```diff
@@ -61,6 +61,51 @@ You can also always find the full command-line options that are available with
...
+   `$ llvm-profdata merge -sparse run1.profraw run2.profraw -o runs.profdata`
+
+4. Visualize the coverage data in HTML:
+
+   `$ llvm-cov show target/.../compiler --format=html -instr-profile=runs.profdata > index.html`
+   
+   There are many visualization and coverage-report options available (see `llvm-cov show --help`).
+
+Note:
+- We recommend using LLVM 11 and a recent nightly version of the Rust toolchain.
```
nagisa, line 103: Maybe "at least LLVM 11" would be better? The tools shipped with LLVM 12 will most likely do just as well if not better than the LLVM11 ones.

README.md
```diff
@@ -61,6 +61,51 @@ You can also always find the full command-line options that are available with
...
+
+   `$ llvm-cov show target/.../compiler --format=html -instr-profile=runs.profdata > index.html`
+   
+   There are many visualization and coverage-report options available (see `llvm-cov show --help`).
+
+Note:
+- We recommend using LLVM 11 and a recent nightly version of the Rust toolchain. 
+  This code was tested with `1.51.0-nightly (2021-02-10)`.
+- Coverage information will be written to the `.profraw` file after `cargo fuzz` successfully finishes running.
+  One way to ensure this is to use the `--max_total_time=<seconds>` or `--runs=<number>` libfuzzer option,
```
nagisa, line 106: Is "this" in "ensure this" the successful termination? Would fuzzer finding a crash count as a successful termination?

src/project.rs
```diff
@@ -160,10 +160,23 @@ impl FuzzProject {
...
-                                     -Cllvm-args=-sanitizer-coverage-pc-table \
                                      -Clink-dead-code"
             .to_owned();
+
+        match build.coverage_output_file {
+            Some(_) => {
+                rustflags.push_str(" -Zinstrument-coverage");
+            }
+            None => {
+                // These two flags currently cause a linking error when combined with `-Zinstrument-coverage`,
```
nagisa, line 171: Is there a bug that can be referenced here? It seems fairly unfortunate to have to tone down sanitization when the coverage is enabled. There is no obvious indication that one should not enable the `--coverage` option, for instance, and ideally there wouldn't be any need for it.

src/project.rs
```diff
@@ -229,6 +242,16 @@ impl FuzzProject {
         Ok(cmd)
     }
 
+    fn with_coverage_output_extension(filename: &str) -> Result<String> {
+        let file_with_ext = Path::new(filename).with_extension("profraw");
+        match file_with_ext.to_str() {
+            Some(name) => Ok(String::from(name)),
+            None => {
+                bail!("File name for generated coverage output must contain only UTF-8 characters.")
```
nagisa, line 250: You can pass `Path`/`PathBuf` to `env` directly, avoiding the conversion back to `String` and thus this failure mode.

nagisa (commented):
Pretty nice!

README.md
amaurremi: I completely agree; however, this would require starting a coverage-report-generation section in the book and it might take some time to do. I suggest to keep it here until one of us finds the time to do that?

src/project.rs
amaurremi: I agree that removing those flags significantly affects the fuzzing, so I put them back. They do cause linking errors when fuzzing some larger programs. It's been difficult to minimize those examples into reportable test cases but I hope to find the time to try.

mykter:
The approach often taken to understanding coverage of fuzzing runs is to record the coverage hit by running through all of the entries in the corpus directory, rather than whilst doing actual fuzzing. The corpus directory contains all of the test cases required to understand the total coverage of the fuzzing run so far.

What do you think about changing this from an option to run, to its own command? e.g. `cargo fuzz coverage <target> <optional corpus dir>`

By doing this, we don't impact normal fuzzing at all, it is faster to run, and easier to use (no question of when to stop).

README.md
fitzgen: FWIW, I'm fine having this here for now.

src/options.rs
```diff
@@ -113,6 +113,11 @@ pub struct BuildOptions {
     #[structopt(long = "target-dir")]
     /// Target dir option to pass to cargo build.
     pub target_dir: Option<String>,
+
+    #[structopt(long = "coverage")]
+    /// Instrument program code with source-based code coverage information.
+    /// The profiler data will be saved in the given file.
+    pub coverage_output_file: Option<String>,
```
fitzgen, line 120: It would be great if the average user didn't have to think about what to name the profile data file. Is there an existing convention that we can use by default?

tests/tests/main.rs
```diff
@@ -243,6 +243,40 @@ fn run_with_crash() {
...
+        .arg("--coverage")
+        .arg("coverage_output")
+        .arg("with_coverage")
+        .arg("--")
+        .arg("-runs=100")
+        .assert()
+        .stderr(
+            predicate::str::contains("Done 100 runs").and(predicate::str::contains(format!(
+                "Raw coverage data saved in coverage_output.profraw"
+            ))),
```
fitzgen, line 275: Can we also assert that the `coverage_output.profraw` file exists?

fitzgen (commented):
Thanks for all this @amaurremi! This is a highly desired feature, and you'll make a lot of people happy by implementing this!

I also think that a new subcommand rather than adding a flag to the existing `cargo fuzz run` subcommand (i.e. `cargo fuzz coverage ...` rather than `cargo fuzz run --coverage ...`) is the way to go here.

src/options.rs
mykter: clang uses `default.profraw` if not told otherwise:

> When the program exits it will write a raw profile to the path specified by the LLVM_PROFILE_FILE environment variable. If that variable does not exist, the profile is written to default.profraw in the current directory of the program. 

https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#running-the-instrumented-program

amaurremi: I am now generating `default-<corpus file name>.profdata` files for each individual input file from the corpus, and merging them in the end using the profdata tool. The user doesn't need to provide names for the raw files.

tests/tests/main.rs
amaurremi: Yes, that makes sense. I now added that check right after we generate that file (i.e. outside of the test).

amaurremi: (And also to the test.)

README.md
```diff
@@ -61,6 +65,51 @@ You can also always find the full command-line options that are available with
...
+
+### Example
+
+Suppose we have a `compiler` fuzz target for which we want to visualize code coverage.
+
+1. Run the fuzzer on the `compiler_fuzzer` target:
+
+   `$ cargo fuzz run compiler_fuzzer`
+
+2. Visualize the coverage data in HTML:
```
mykter, line 102: Now the approach has changed we need a new step, running cargo fuzz coverage.

README.md
```diff
@@ -61,6 +65,51 @@ You can also always find the full command-line options that are available with
...
+
+After you fuzzed your program, use the `coverage` command to generate precise
+[source-based code coverage](https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html)
+information:
+```
+$ cargo fuzz coverage <target> [corpus dir]
+```
+This command
+
+- compiles your project using the `-Zinstrument-coverage` Rust compiler flag,
```
mykter, line 86: Are there any issues with running `fuzz coverage` whilst fuzzing is ongoing (e.g. in another terminal)?
Similarly with alternating between fuzz run and fuzz coverage?

If it potentially causes issues it would be helpful to document them (if we can't work around them with, say, separate target binaries for coverage-instrumented vs fuzzing-instrumented).

amaurremi: I tried running `cargo fuzz run` and `cargo fuzz coverage` on the same target (a WebAssembly interpreter) in parallel and didn't encounter issues. The `coverage` command generated information for the corpus data that was available at the moment it was run, and I was able to re-run it while the fuzzer was still running. I was also able to re-run the fuzzer while the coverage information was being produced. 
However, note that alternating between the `run` and `fuzz` commands involves rebuilding the target because they require different compiler options.
Are there more potential issues you would be looking for? I imagine that in general one would want to avoid running the two modes in parallel to maximize the computing resources available to the fuzzer.

mykter: > I was also able to re-run the fuzzer while the coverage information was being produced.

I don't fully understand how this works, but if it does then that's great!

My confusion is that there is only a single target binary path, so it seems we either have a binary with coverage instrumentation or not, and will never have both at the same time. In the scenario you tested, how do we know which version is going to get run? Might the coverage loop switch to the non-coverage-instrumented binary half way through, when it gets compiled for `cargo fuzz run`?

src/options.rs
```diff
@@ -113,6 +115,11 @@ pub struct BuildOptions {
     #[structopt(long = "target-dir")]
     /// Target dir option to pass to cargo build.
     pub target_dir: Option<String>,
+
+    #[structopt(skip = false)]
```
fitzgen, line 119: I think the `skip = false` bit isn't necessary, is it? By default structopt will not skip fields.

Or is this trying to make this option skipped, so that it is only exposed with `cargo fuzz coverage`?

src/options/coverage.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{options::BuildOptions, project::FuzzProject, RunCommand};
+use anyhow::Result;
+use structopt::StructOpt;
+
+#[derive(Clone, Debug, StructOpt)]
+pub struct Coverage {
+    #[structopt(flatten)]
+    pub build: BuildOptions,
+
+    #[structopt(required(true))]
```
fitzgen, line 10: Similarly, I think that non-`Option`/non-`Vec` fields are required by default, so this `required(true)` annotation shouldn't be necessary here.

src/project.rs
```diff
@@ -572,6 +577,105 @@ impl FuzzProject {
...
+        }
+
+        let (coverage_out_raw_dir, coverage_out_file) = self.coverage_for(&coverage.target)?;
+
+        // Generating individual coverage data for all files in corpora
+        for input_file in readable_input_files {
+            let (mut cmd, file_name) =
+                self.create_coverage_cmd(coverage, &coverage_out_raw_dir, &input_file.path())?;
+            eprintln!("Generating coverage data for {:?}", file_name);
+            cmd.output()
```
fitzgen, line 620: Calling `.output()` and then not using the output is a bit inefficient. `.output()` will buffer up stdout and stderr and return them, but if we don't actually care about the output, and we also don't want it printed to *our* stderr/stdout, then we can do 

```rust
cmd
    .stdout(std::process::Stdio::null())
    .stderr(std::process::Stdio::null())
```

when constructing the command, and then run the command via

```rust
let status = cmd.status().with_context(|| ...)?;
anyhow::ensure!(status.success, "command exited with failure code {}: {:?}", status.code, cmd);
```

(Note: we also want to check that the command exited successfully, as the above snippet does)

fitzgen (commented):
Nice! A couple nitpicks to fix before we merge this but, once fixed, I think this is good to go! Thanks again @amaurremi!

src/options.rs
amaurremi: > Or is this trying to make this option skipped, so that it is only exposed with cargo fuzz coverage?

Yes, the syntax here is `skip = <default value>`, which says that the option should not be shown to the user and specifies a default value. In our case the default value is `false`. I agree that this looks confusing.

My intention was to avoid exposing this flag to the user. It's a build flag that should always be used with the `coverage` command, and the code ensures that this flag is set with that command.

src/options/coverage.rs
amaurremi: Ah, OK. Could you help me understand how this case is different from similar instances in options for the other command? For example, [here](https://github.com/rust-fuzz/cargo-fuzz/blob/master/src/options/run.rs#L10).

fitzgen: Unless I'm mistaken, those uses of `required(true)` can also be removed.

src/options.rs
fitzgen: Ah okay, thanks for the explanation. Would you mind explaining this a bit more in the comment? Thanks.

README.md
amaurremi: That's a good point. I decided to just in case create a separate build target for `cargo fuzz coverage`.

src/project.rs
amaurremi: I actually do want to preserve the output of those commands, especially if there is a failure. So I am just using `cmd.status...` similarly to what you suggested.

amaurremi:
Thank you for all the feedback. Always happy to receive more suggestions if you have any.

src/options/tmin.rs
```diff
@@ -8,7 +8,6 @@ pub struct Tmin {
     #[structopt(flatten)]
     pub build: BuildOptions,
 
-    #[structopt(required(true))]
```
fitzgen, line 11: Thanks for cleaning up all these other options!

fitzgen (approved):
Thanks for keeping with the pull request! Looks great, going to merge it now.

PR #249, opened 2021-02-20
Allow passing extra arguments to tmin/cmin

tilpner:
I noticed I needed to pass `-detect_leaks=0 -rss_limit_mb=8192` to `run`, but the CLI didn't allow passing those to `cmin` or `tmin`. So I changed that, and now cmin/tmin seem to do something.


frewsxcv (approved):
ty!

PR #257, opened 2021-03-26
Support rustflags from cargo config

paholg:
Cargo will ignore any rustflags set in a config file if `RUSTFLAGS` is
set. Because cargo-fuzz always sets `RUSTFLAGS`, flags in config files
are always ignored.

So, instead of just pulling prior flags from `RUSTFLAGS`, we can rely on
cargo to find them for us.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Released YYYY-MM-DD.
 
 ### Added
 
-* TODO (or remove section if none)
+* Added support for rustflags set in a cargo config file.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


paholg:
The build failure appears to be a transient network error.

fitzgen:
I *really* don't want to pull in `cargo` as a library. It's a ton of dev overhead; really slow to compile and we'll be chasing updates every 6 weeks. I talked with @alexcrichton and he confirmed that using `cargo` as a library is discouraged, and that instead we should file an issue on the `cargo` repo to add a CLI flag to `cargo` to print out the information we need, and then we can just spawn a subprocess to get this information.

Can you file a bug upstream requesting a CLI flag to read the `rustflags` from the cargo config, and then rewrite this PR to use a `cargo` sub-process? Thanks!

paholg:
Fair enough, will do.

paholg:
It turns out this was just recently implemented in cargo via `cargo config`, so I just need to wait for it to stabilize.

https://github.com/rust-lang/cargo/pull/9302

Issue #254, opened 2021-03-15
Merging raw coverage files fails with "Argument list too long" on macOS


jrmuizel:
I have a corpus of about 3483 different samples. Merging fails when running `llvm-profdata` because it lists each profraw file in the arguments. I believe `llvm-profdata` supports specifying a directory instead of individual file names. Doing that would probably solve this.

fitzgen:
Ah that's no good!

I don't personally have the cycles to fix this right now, maybe @amaurremi does?

silvergasp:
@jrmuizel Does #336 fix your issue?

You can check by running `cargo install --git https://github.com/silvergasp/cargo-fuzz.git cargo-fuzz`

PR #258, opened 2021-03-29
Supplying raw-output-coverage directory to profdata command

amaurremi:
…instead of individual files. This addresses #254.


fitzgen (approved):
Thank you!

PR #260, opened 2021-04-16
Allow stripping of dead code

paholg:
The `-Clink-dead-code` flag was added to fix an error with certain
targets.

However, it can also cause problems.

Example: You depend on crate A which links to library B. A includes an
unused reference to `B::foo`. If the specific version of B we have is
missing `foo`, we now get a linker error whereas we would not if
stripping dead code.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -123,6 +123,11 @@ pub struct BuildOptions {
     /// The attribute takes a default value `false`, ensuring that by default,
     /// the coverage option will be disabled).
     pub coverage: bool,
+
+    /// Dead code is linked by default to prevent a potential error with some
+    /// optimized targets. This flag allows you to opt out of it.
+    #[structopt(long)]
+    pub strip_dead_code: bool,
 }
 
 impl stdfmt::Display for BuildOptions {
@@ -200,6 +205,7 @@ mod test {
             unstable_flags: Vec::new(),
             target_dir: None,
             coverage: false,
+            strip_dead_code: false,
         };
 
         let opts = vec![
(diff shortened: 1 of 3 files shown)


src/options.rs
```diff
@@ -123,6 +123,11 @@ pub struct BuildOptions {
     /// The attribute takes a default value `false`, ensuring that by default,
     /// the coverage option will be disabled).
     pub coverage: bool,
+
+    /// Dead code is linked by default to prevent a potential error with some
+    /// optimized targets. This flag allows you to opt out of it.
+    #[structopt(long = "strip-dead-code")]
```
fitzgen, line 129: nitpick: `structopt` should infer this `long` name, so this line shouldn't be necessary here.

fitzgen (commented):
Thanks! Can you add a small test that passes the `--strip-dead-code` flag to a simple "hello world" style fuzz target just so that we kick the tires on this a bit and hopefully catch any future regressions during, say, a refactoring that break this flag?

src/options.rs
paholg: I am not clear why, but without specifying the name, the `run_with_crash` test fails, but only when it's run as part of the full test suite.

paholg: Ah, I figured out the issue I was running into. Because the test ids (and therefore directories) are non-deterministic, on repeated runs, I the `run_*_with_crash` tests can end up looking at the wrong output.

Running `cargo clean` between tests is a workaround, but I'll open another PR to make the project directories deterministic, as the failures are very confusing.

fitzgen (approved):
Thanks!

paholg:
@fitzgen Thanks for the quick review / merge!

If it's not too much trouble, would you mind doing a version bump / publish soon?

PR #271, opened 2021-08-05
Don't hardcode target strings

alexcrichton:
This should help get `cargo fuzz` working on other systems, such as
aarch64-unknown-linux-gnu, without having to always pass `--target`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,6 @@
+use std::env;
+
+fn main() {
+    println!("cargo:rustc-env=TARGET={}", env::var("TARGET").unwrap());
+    println!("cargo:rerun-if-changed=build.rs");
+}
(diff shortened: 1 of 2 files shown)


Manishearth (approved):

fitzgen:
@alexcrichton do you need a new release with this?

alexcrichton:
Nah not urgent, just got annoyed typing `--target` so often and figured I'd send a PR

build.rs
```diff
@@ -0,0 +1,6 @@
+use std::env;
+
+fn main() {
+    println!("cargo:rustc-env=TARGET={}", env::var("TARGET").unwrap());
+    println!("cargo:rerun-if-changed=build.rs");
```
Shnatsel, line 5: I think this line will break cross-compilation: if I first compile for e.g. linux-gnu and then linux-musl (or Mac or whatever), the TARGET variable will still be set to linux-gnu. Or am I missing something?

Shnatsel: Sounds like `println!("cargo:rerun-if-changed-env=TARGET");` should work, but I'm not sure if it's the case for the variables that Cargo sets internally

Shnatsel: Yep, I've tested and it works. I'll make a separate PR.

PR #277, opened 2021-09-26
Adjust for NewPM

### LINKED ISSUE #276: Incompatible with nightly-2021-09-26: unknown pass name 'sancov'
I noticed that serde_json's fuzz target does not build using the most recent nightly. It's been working for a long time so this is a recent breakage.

Repro:

```console
$ cargo fuzz init
$ cargo +nightly-2021-09-25 fuzz build  # works
$ cargo +nightly-2021-09-26 fuzz build

error: failed to run LLVM passes: unknown pass name 'sancov'

error: could not compile `once_cell` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
Error: failed to build fuzz script: "cargo" "build" "--manifest-path" "fuzz/Cargo.toml" "--target" "x86_64-unknown-linux-gnu" "--release" "--bins"
```

The upstream commit range is https://github.com/rust-lang/rust/compare/73422130ee96c09e7214c876a3600ac1f32aa8c8...addb4da686a97da46159f0123cb6cdc2ce3d7fdb.

nagisa:
This is necessary to work with the most recent rust nightlies. As a
downside, cargo-fuzz will stop working with the older versions of
nightly...

Fixes #276

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,19 +5,24 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Check rustfmt
-      run: |
-        rustup component add rustfmt --toolchain stable
-        cargo +stable fmt --all -- --check
-    - name: Build
-      run: cargo build --verbose
-    - name: Run tests
-      run: |
-        rustup install nightly
-        rustup default nightly
-        rustup component add rust-src
-        rustup component add llvm-tools-preview
-        cargo install cargo-binutils
-        export RUST_BACKTRACE=1
-        cargo test --verbose
+    - uses: actions/checkout@v2
+    - uses: actions-rs/toolchain@v1
+      with:
+          toolchain: nightly
+          profile: minimal
+          default: true
+          components: rustfmt, rust-src, llvm-tools-preview
+    - uses: actions-rs/cargo@v1
+      with:
+        command: fmt
+        args: --all -- --check
+    - uses: actions-rs/cargo@v1
+      with:
+        command: install
+        args: cargo-binutils
+    - uses: actions-rs/cargo@v1
+      with:
+        command: test
+        args: --verbose
+      env:
+        RUST_BACKTRACE: 1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -111,7 +111,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-fuzz"
-version = "0.10.2"
+version = "0.11.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-fuzz"
-version = "0.10.2"
+version = "0.11.0"
 authors = ["The rust-fuzz Project Developers"]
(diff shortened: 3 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -5,19 +5,23 @@ jobs:
...
+    - uses: actions/checkout@v2
+    - uses: actions-rs/toolchain@v1
+      with:
+          toolchain: nightly
+          profile: minimal
+          default: true
+          components: rustfmt, rust-src, llvm-tools-preview
+    - uses: actions-rs/cargo@v1
+      with:
+        command: fmt
```
alex, line 17: This is missing
```
          args: --all -- --check
```

I think

alex (commented):
FWIW in the future I think it'd be better to move the CI cleanup to a separate PR for ease of reviewability.

Do we document our minimum nightly version anywhere? This obviously increases it to a very recent nightly.

nagisa:
We don't document specific nightly bounds, but we do require a nightly and as far as I'm concerned that means the last available one.

alex (approved):
good enough for me

Manishearth:
Can we do some kind of version checking so we don't drop support for older compilers?

nagisa:
I don't want to spend my effort on that, but yeah it would be possible. We
can just test for this by attempting to compile a test program.

alex:
My personal vote would be to land + release as is so there's something out there that works with the latest nightly (selfishly, since this impacts one of my projects) and then we can do a follow-up if someone is motivated to implement it.

Manishearth:
@nagisa hmm. I was just thinking of parsing `rustc --version`. I guess this is fine, but also I wonder if we can get rustc to have a transition period for this change where there's a compat mode for a while.

nagisa:
`rustc --version` is more difficult and less accurate. It won't capture cases where `rustc` is built against system LLVM that does not support newPM etc.

There is the `-Znew-llvm-pass-manager=no` which allows reverting back to the old PM for builds here. This would allow supporting a broader sweep of rustc versions, but IMO its a bad idea to have so significantly different build configurations between regularly built code (which uses newpm by default) and fuzz code.

Manishearth:
> There is the `-Znew-llvm-pass-manager=no` flag which allows reverting back to the old PM for builds here. This would allow supporting a broader sweep of rustc versions, but IMO its a bad idea to have so significantly different build configurations between regularly built code (which uses newpm by default) and fuzz code.

No, I was suggesting that we get rustc to silently translate `-Cpasses=sanconv` to `-Cpasses=sanconv-module` until such a time when we can have a better compat story here.

fitzgen:
> My personal vote would be to land + release as is so there's something out there that works with the latest nightly (selfishly, since this impacts one of my projects) and then we can do a follow-up if someone is motivated to implement it.

Agreed. Let's fix the breakage and then add the backwards compat in follow ups.

fitzgen:
Published.

matklad:
For fellow goolerers: if you see 

```
ERROR: no interesting inputs were found. Is the code instrumented for coverage? Exiting.
```

and 

```
warning: unknown pass `sancov-module`, ignoring
```

the issue is that you are using new `cargo fuzz` with old `rustc`. You might want to `cargo install cargo-rustc --vers 0.10.0` to downgrade `cargo-fuzz` or to update rustc version.

PR #279, opened 2021-09-29
Include the fuzzer entrypoint on Windows

danielframpton:
The current entrypoint for fuzzing ends up in a bundled static library in an rlib.

This change tells the linker to include main, which will then cause the entrypoint to be found and included from the rlib.

Note that this also requires a change on the rust side https://github.com/rust-lang/rust/pull/89369 to enable sanitizer support.

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -196,6 +196,10 @@ impl FuzzProject {
         if !build.release || build.debug_assertions {
             rustflags.push_str(" -Cdebug-assertions");
         }
+        if build.triple.contains("-msvc") {
+            // The entrypoint is in the bundled libfuzzer rlib, this gets the linker to find it.
+            rustflags.push_str(" -Clink-arg=/include:main");
+        }
 
         // If release mode is enabled then we force 1 CGU to be used in rustc.
         // This will result in slower compilations but it looks like the sancov


nagisa:
This seems like a fairly unfortunate workaround to have. Is there an explanation for why `link.exe` discards `main` here? I mildly remember Windows expecting a different entrypoint name, maybe we could introduce an alias or somesuch?

danielframpton:
It is not so much that link discards main, but that it never loaded the object with main in the first place (it doesn't automatically get included on the list of required symbols). It is only after doing a pass of linking that link then searches for `main` or `wmain` (or others like `WinMain` for different subsystems) to infer the correct crt entrypoint to use based on the symbols that it has seen. So I don't think we can alias anything, we need to do something to get the object that has `main` linked in.

Since in this case we know the method we need to include from the fuzzing library `/include:main` seemed like a straightforward way to require the linker to include it and let the rest of the logic flow from there.

Another option (I believe clang took this approach) would be to have the fuzzing library be passed as whole archive to link, but that would preclude it from being bundled in the rlib.

We could also explicitly specify the entrypoint (e.g., `/entry:mainCRTStartup`) but that brings in more dependency/knowledge of something else that doesn't seem necessary so I think it is cleaner to only rely on the knowledge that `main` is the post-crt entry point for libfuzzer and get that object included.

danielframpton:
Thanks @nagisa!

nagisa:
Since the Rust side of the functionality is not yet available, we'll wait to release a new version, but let us know via an issue or a comment if you'd like one to be released ASAP.

PR #288, opened 2022-01-24
Add --no-trace-compares flag

### LINKED ISSUE #287: Targets built with cargo-fuzz always require POPCNT instruction
libFuzzer uses the `POPCNT` instruction for data-flow guided coverage and cargo-fuzz builds targets with it turned on. This causes the targets to mysteriously crash when the CPU doesn't support the instruction (like my ancient laptop). References:

- [cargo-fuzz enabling `-sanitizer-coverage-trace-compares`](https://github.com/rust-fuzz/cargo-fuzz/blob/63730da7f95cfb21f6f5a9b0a74532f98d3983a4/src/project.rs#L164)
- [Various libFuzzer functions in `FuzzerTracePC.cpp` with an attribute to always emit `POPCNT`](https://github.com/rust-fuzz/libfuzzer/blob/fcf3b18c854e2e5e7f61a8a701f72ef051b1c7e1/libfuzzer/FuzzerTracePC.cpp)

What do you think about adding a flag to turn off any `-sanitizer-coverage-trace-*` option that cargo-fuzz adds by default?

LynnKirby:
The flag will opt-out of the `-sanitizer-coverage-trace-compares` LLVM flag that cargo-fuzz builds with by default. The LLVM flag can help with fuzzing but it's not without overhead so disabling it can improve performance. It also allows older CPUs without the `popcnt` instruction to run the binaries built by cargo-fuzz since the `trace-compares` instrumentation requires it.

Fixes #287.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ Released YYYY-MM-DD.
 * Added the `--no-trace-compares` flag which opts out of the
   `-sanitizer-coverage-trace-compares` LLVM argument.
 
-  Using this may improve performance at the cost of worse fuzzing accuracy.
-  It also allows older CPUs lacking the `popcnt` instruction to use cargo-fuzz;
+  Using this may improve fuzzer throughput at the cost of worse coverage accuracy.
+  It also allows older CPUs lacking the `popcnt` instruction to use `cargo-fuzz`;
   the `*-trace-compares` instrumentation assumes that the instruction is
   available.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,13 @@ Released YYYY-MM-DD.
 
 ### Added
 
-* TODO (or remove section if none)
+* Added the `--no-trace-compares` flag which opts out of the
+  `-sanitizer-coverage-trace-compares` LLVM argument.
+
+  Using this may improve performance at the cost of worse fuzzing accuracy.
+  It also allows older CPUs lacking the `popcnt` instruction to use cargo-fuzz;
+  the `*-trace-compares` instrumentation assumes that the instruction is
+  available.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,13 @@ Released YYYY-MM-DD.
...
 ### Added
 
-* TODO (or remove section if none)
+* Added the `--no-trace-compares` flag which opts out of the
+  `-sanitizer-coverage-trace-compares` LLVM argument.
+
+  Using this may improve performance at the cost of worse fuzzing accuracy.
+  It also allows older CPUs lacking the `popcnt` instruction to use cargo-fuzz;
+  the `*-trace-compares` instrumentation assumes that the instruction is
+  available.
```
fitzgen, line 13:
```suggestion
  Using this may improve fuzzer throughput at the cost of worse coverage accuracy.
  It also allows older CPUs lacking the `popcnt` instruction to use `cargo-fuzz`;
  the `*-trace-compares` instrumentation assumes that the instruction is
  available.
```

src/options.rs
```diff
@@ -133,6 +133,10 @@ pub struct BuildOptions {
     /// allows you to opt out of it.
     #[structopt(long)]
     pub no_cfg_fuzzing: bool,
+
+    #[structopt(long)]
+    /// Don't build with the `sanitizer-coverage-trace-compares` LLVM argument
```
fitzgen, line 138:
```suggestion
    /// Don't build with the `sanitizer-coverage-trace-compares` LLVM argument
    ///
    ///  Using this may improve fuzzer throughput at the cost of worse coverage accuracy.
    /// It also allows older CPUs lacking the `popcnt` instruction to use `cargo-fuzz`;
    /// the `*-trace-compares` instrumentation assumes that the instruction is
    /// available.
```

fitzgen (commented):
Thanks! Just a couple nitpicks below.

PR #294, opened 2022-03-10
Switch `-Zinstrument-coverage` to `-C`

alexcrichton:
This is a stable option now so it can use `-C` to prevent a warning from
being emitted from rustc


Manishearth (approved):

fitzgen:
Thanks!

PR #298, opened 2022-04-13
Use the `current_platform` crate instead of shipping with a copy-paste of it

Shnatsel:
Drops the custom build.rs and frees it up for other things.

The `rerun-if-changed-env` directive that's currently in use, while necessary, [disables the default change detection heuristics](https://github.com/rust-lang/cargo/issues/4587). This makes it difficult to use build.rs in the future for anything other than re-exporting the variable.

The `current_platform` crate used instead is maintained by me and is basically #296 in crate form.

I have no strong feelings for or against this, so feel free to turn this down.


Shnatsel:
I'm not sure what's up with CI - `cargo test` passes for me locally.

Shnatsel:
Looks like CI is broken: #299 only changes a single string, no code, and also fails.

fitzgen (approved):
PR looks good to me.

Unsure what is up with CI. It looks like the expected MSan crash report isn't happening and instead there is just a deadly signal?

In any case, I think it is probably fine to merge this now without fixing CI since the crate still builds and can run fuzz targets and all that.

fitzgen:
We probably shouldn't cut a new release until we fix CI tho, fwiw.

PR #317, opened 2022-07-30
Add `check` subcommand to perform a `cargo check`

dtolnay:
`cargo fuzz check`. Useful for CI where you aren't going to **run** the fuzz target, only verify that it compiles. (Running is left to OSS-Fuzz.)


dtolnay:
CI failure is the same as #316, unrelated to the PR as far as I can tell.

PR #327, opened 2022-12-12
Don't rebuild coverage target on each coverage run

sharnoff:
Aside from reducing the amount of output from repeated `cargo run` executions (particularly when there's warnings), this also reduced the total time to run `cargo fuzz coverage` in my project from ~15 to ~4 seconds.

The new behavior should still be safe from other code changes in the project being fuzzed  while running (and other `cargo fuzz` invocations), because the coverage directory is kept separate.

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -266,6 +266,36 @@ impl FuzzProject {
         Ok(cmd)
     }
 
+    fn target_dir(&self, build: &options::BuildOptions) -> Result<PathBuf> {
+        if build.coverage {
+            Ok(env::current_dir()?
+                .join("target")
+                .join(default_target())
+                .join("coverage"))
+        } else {
+            Ok(build
+                .target_dir
+                .as_ref()
+                .map(PathBuf::from)
+                .unwrap_or_else(|| self.project_dir.join("target")))
+        }
+    }
+
+    fn cargo_build_bin_path(
+        &self,
+        build: &options::BuildOptions,
+        fuzz_target: &str,
+    ) -> Result<PathBuf> {
+        let profile_subdir = if build.dev { "debug" } else { "release" };
+
+        let bin_path = self
+            .target_dir(&build)?
+            .join(&build.triple)
+            .join(profile_subdir)
+            .join(fuzz_target);
+        Ok(bin_path)
+    }
+
     pub fn exec_build(
         &self,
         mode: options::BuildMode,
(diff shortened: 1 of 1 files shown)


src/project.rs
```diff
@@ -266,6 +266,36 @@ impl FuzzProject {
...
+            Ok(env::current_dir()?
+                .join("target")
+                .join(default_target())
+                .join("coverage"))
+        } else {
+            Ok(build
+                .target_dir
+                .as_ref()
+                .map(PathBuf::from)
+                .unwrap_or_else(|| self.project_dir.join("target")))
```
fitzgen, line 280: If the user explicitly passes in a `--target-dir` I think that should take priority over the default coverage target directory. So I think this should look something like:

```rust
if let Some(target_dir) = build.target_dir {
    // use user-provided target dir...
} else if build.coverage {
    // use coverage target dir...
} else {
    // use default target dir...
}
```

src/project.rs
```diff
@@ -284,15 +314,8 @@ impl FuzzProject {
...
-            cmd.arg("--target-dir").arg(target_dir);
-        } else if build.coverage {
-            // To ensure that fuzzing and coverage-output generation can run in parallel, we
-            // produce a separate binary for the coverage command.
-            let target_dir = env::current_dir()?
-                .join("target")
-                .join(default_target())
-                .join("coverage");
+        if build.target_dir.is_some() || build.coverage {
+            let target_dir = self.target_dir(&build)?;
```
fitzgen, line 318: Can we make the `target_dir` method return an `Option` and do `if let Some(target) = self.target_dir(&build)` here? That way we are less likely to specify the `--target-dir` flag to `cargo` when we intend to use the default (and the default might have been configured to a non-default location by the user in `.cargo/config.toml` or via env vars so it is best to omit the `--target-dir` flag than to guess its location by appending `<project>/target` or something like that)

src/project.rs
fitzgen, line 280: `self.project_dir.join("target")` is going to be incorrect if the user using either the `CARGO_TARGET_DIR=...` environment variable or configuring it via `.cargo/config.toml`. In this case, we should return `None`.

fitzgen (commented):
Thanks!

sharnoff:
Thanks for the quick review! Let me know if this works :)

fitzgen (approved):
Thanks!!

PR #334, opened 2023-02-07
Run deploy actions with correct tags

Luni-4:
Tags are written without the `v` char, so let's update the deploy actions script


fitzgen (approved):
Thanks!

Luni-4:
Thanks to you @fitzgen! Could you release a new patch version to produce binaries, please?

PR #336, opened 2023-02-15
perf(coverage): Improve coverage collection by using merge

### LINKED ISSUE #254: Merging raw coverage files fails with "Argument list too long" on macOS

silvergasp:
This approach creates a dummy-corpus and merges the main corpus into the dummy-corpus. After the corpus is merged we delete the dummy-corpus. This allows us to calculate code coverage signficantly faster than calculating it per-file in the corpus, as it can be run in parallel and there is less startup initialisation cost for creating a new process for each entry in the corpus.

This was mainly inspired by the approach taken in [google/oss-fuzz](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-runner/coverage#L83-L96).

This should also fix #254 as there will only be one set of profile data per corpus directory rather than a set of profile data per file in all the corpus directories. Although I can't confirm this as I don't own a mac :)

# Performance diff:
Testing performance on the fuzzers in [quick-xml](https://github.com/tafia/quick-xml). With a corpus of 38780 files.

## Before 
```
time cargo fuzz coverage
# ...
real    43m51.644s
user    9m10.208s
sys     11m41.468s
```
## After
```
time cargo fuzz coverage
# ...
real    0m25.826s
user    1m14.378s
sys     0m6.698s
```

## Delta
real: 101x faster

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -671,11 +671,11 @@ impl FuzzProject {
 
         let (coverage_out_raw_dir, coverage_out_file) = self.coverage_for(&coverage.target)?;
 
-        // Generating individual coverage data for all files in corpora.
-        for input_file in readable_input_files {
-            let (mut cmd, file_name) =
-                self.create_coverage_cmd(coverage, &coverage_out_raw_dir, &input_file.path())?;
-            eprintln!("Generating coverage data for {:?}", file_name);
+        for corpus in corpora.iter() {
+            // _tmp_dir is deleted when it goes of of scope.
+            let (mut cmd, _tmp_dir) =
+                self.create_coverage_cmd(coverage, &coverage_out_raw_dir, &corpus.as_path())?;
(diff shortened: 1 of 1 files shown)


silvergasp:
I've also tested this change to ensure that the same coverage is calculated before and after the change.
```
 $ rm fuzz/coverage/fuzz_target_1/coverage.profdata
 $ cargo fuzz coverage -O fuzz_target_1
 $ cargo cov -- report fuzz/target/x86_64-unknown-linux-gnu/release/fuzz_target_1 -instr-profile fuzz/coverage/fuzz_target_1/coverage.profdata
 src/
```

PR #339, opened 2023-05-04
Add --careful to emulate cargo-careful

### LINKED ISSUE #331: Integrate cargo-careful
cargo-careful enables debug assertions and some UB checks in libstd, I think it would be useful if we can fuzz under that configuration
https://github.com/RalfJung/cargo-careful

(@RalfJung Might be of interest to you)

f0rki:
Add the option to enable "careful mode" inspired by @RalfJung 's cargo-careful project. This should fix #331.

* based on the compiler flags from https://github.com/RalfJung/cargo-careful/
* includes PR #292 by @saethlin, but disables defaulting to `-Zbuild-std` - `--careful` does imply `--build-std`.


fitzgen:
Thanks!

elichai:
This is great! Thank you!

PR #340, opened 2023-05-17
Add an option to disable instrumenting Rust code for fuzzing

Voker57:
The reason for this is sometimes I only want to fuzz the specific part of code, and don't want to fuzz the whole Rust crate stack.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -126,6 +126,10 @@ pub struct BuildOptions {
     #[arg(long)]
     pub no_cfg_fuzzing: bool,
 
+    /// Do not instrument Rust code for fuzzing
+    #[arg(long)]
+    pub no_rust_fuzzing: bool,
+
     #[arg(long)]
     /// Don't build with the `sanitizer-coverage-trace-compares` LLVM argument
     ///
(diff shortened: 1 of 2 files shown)


fitzgen:
> I only want to fuzz the specific part of code, and don't want to fuzz the whole Rust crate

This PR would turn off fuzzing instrumentation for *all* code though, including the "specific part of code" so I don't think this is doing what you say you want to do?

Unfortunately, I don't think there is a good way to do what you describe other than writing your fuzz target such that code you don't want to fuzz is statically unreachabel (i.e. your fuzz target can only exercise the code you want to fuzz).

Voker57:
> > I only want to fuzz the specific part of code, and don't want to fuzz the whole Rust crate
> 
> This PR would turn off fuzzing instrumentation for _all_ code though, including the "specific part of code" so I don't think this is doing what you say you want to do?
> 
> Unfortunately, I don't think there is a good way to do what you describe other than writing your fuzz target such that code you don't want to fuzz is statically unreachabel (i.e. your fuzz target can only exercise the code you want to fuzz).

Some code is already instrumented and linked as a static library.

Issue #333, opened 2023-02-06
`cargo fuzz init` fails with `edition.workspace = true`


mauricelam:
When running `cargo fuzz init` on a workspace, where the member crate's Cargo.toml has `edition.workspace = true`, the command fails.

`Cargo.toml`:
```toml
[worksapce]
members = [
  "member",
]

[workspace.package]
edition = "2021"
```

`member/Cargo.toml`:
```toml
[package]
name = "member"
version = "0.1.0"
edition.workspace = true
```

```sh
$ cd member
$ cargo fuzz init
Error: /path/to/project/member/Cargo.toml (package.edition) is malformed
```

`edition.workspace` is supported by cargo itself per https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table. If I change `edition.workspace = true` back to `edition = "2021"`, the init command runs successfully.

exdx:
Running into the same issue -- my first time trying to use the project. The workspace `Cargo.toml` looks like this:
```
[workspace]
members = [
  "a",
  "b",
  "c",
  "d",
  "e",
]

[profile.release]
debug = true
```

I think based on https://github.com/rust-fuzz/cargo-fuzz/issues/226#issuecomment-629822514 if you run `cargo fuzz init` inside a particular workspace member it works. Worked fine for me. Maybe the docs should be updated to make that clear in the case of workspaces.

mauricelam:
@exdx That sounds like a different problem? I was running `cargo fuzz init` from inside the `member` directory but it still fails if the member has `edition.workspace = true`.

PR #349, opened 2023-12-08
fix: use cargo-metadata to support init command in workspace

0xalpharush:
closes https://github.com/rust-fuzz/cargo-fuzz/issues/333

Instead of manually parsing the toml file and failing in workspace members, the package name and edition are obtained via cargo-metadata (this also seems to obviate the need to pass that path which is a breaking change to a public API). The way the tests are setup was making cargo-metadata think that each test is a member of cargo-fuzz so I added an empty `workspace` entry to the tests' Cargo.toml.

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -932,7 +932,12 @@ pub struct Manifest {
 impl Manifest {
     pub fn parse() -> Result<Self> {
         let metatdata = MetadataCommand::new().exec()?;
-        let package = metatdata.packages.first().expect("at least one package");
+        let package = metatdata.packages.first().with_context(|| {
+            anyhow!(
+                "Expected to find at least one package in {}",
+                metatdata.target_directory
+            )
+        })?;
         let crate_name = package.name.clone();
         let edition = Some(String::from(package.edition.as_str()));
 
--- a/src/project.rs
+++ b/src/project.rs
@@ -53,8 +53,7 @@ impl FuzzProject {
     pub fn init(init: &options::Init, fuzz_dir_opt: Option<PathBuf>) -> Result<Self> {
         let project = Self::manage_initial_instance(fuzz_dir_opt)?;
         let fuzz_project = project.fuzz_dir();
-        let root_project_manifest_path = project.project_dir.join("Cargo.toml");
-        let manifest = Manifest::parse(&root_project_manifest_path)?;
+        let manifest = Manifest::parse()?;
 
         // TODO: check if the project is already initialized
         fs::create_dir(fuzz_project)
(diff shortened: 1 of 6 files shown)


fitzgen:
I've had issues with `cargo-metadata` blowing up compile times in the past. Can you provide the timings for `cargo clean; cargo build` both on `main` and on this feature branch?

0xalpharush:
Using hyperfine on my M1, it adds ~1s

Feature branch:
```
Benchmark 1: cargo clean; cargo build
  Time (mean ± σ):      6.016 s ±  0.043 s    [User: 21.272 s, System: 2.145 s]
  Range (min … max):    5.968 s …  6.093 s    10 runs
```
 Main branch:
 ```
 Benchmark 1: cargo clean; cargo build
  Time (mean ± σ):      5.009 s ±  0.113 s    [User: 14.871 s, System: 1.736 s]
  Range (min … max):    4.859 s …  5.199 s    10 runs
  ```

src/project.rs
```diff
@@ -931,25 +932,18 @@ pub struct Manifest {
...
-            .transpose()?;
+        let metatdata = MetadataCommand::new().exec()?;
+        let package = metatdata
+            .packages
+            .iter()
+            .find(|p| p.manifest_path == path)
+            .expect({
+                let path = path.display();
+                &format!("could not find package for {}", path)
+            });
```
fitzgen: Rather than panic (via `expect`) can we return actual errors? That will also fix the issue of always allocating the panic message string even when we don't panic here.

src/project.rs
```diff
@@ -931,16 +930,9 @@ pub struct Manifest {
...
         let metatdata = MetadataCommand::new().exec()?;
-        let package = metatdata
-            .packages
-            .iter()
-            .find(|p| p.manifest_path == path)
-            .expect({
-                let path = path.display();
-                &format!("could not find package for {}", path)
-            });
+        let package = metatdata.packages.first().expect("at least one package");
```
fitzgen, line 935: Again, it isn't clear to me that this is something we can `expect` and shouldn't gracefully produce and propagate an error when we run into?

fitzgen (commented):
One second doesn't seem too bad, given that this fixes bugs and cleans up some fiddly bits.

Some questions below.

src/project.rs
0xalpharush: https://github.com/rust-fuzz/cargo-fuzz/pull/349/commits/3a91261ea379943c256ea022f83e418a614a634f

src/project.rs
0xalpharush: This was removed in https://github.com/rust-fuzz/cargo-fuzz/pull/349/commits/a4a946e8fd4f945f042691b23f5a04d7f34c98bd

fitzgen (approved):
Thanks!

Manishearth (approved):

0xalpharush:
The tests were relying on the panic error format pre rustc 1.73.0 https://blog.rust-lang.org/2023/10/05/Rust-1.73.0.html

fitzgen (approved):
Thanks!

PR #354, opened 2024-01-02
add llvm-path option

### LINKED ISSUE #353: Add option to provide `llvm-path` in `cargo fuzz coverage`
It would be good if you could provide cargo-fuzz with the location of an already installed instance of `llvm-profdata` to run on `cargo fuzz coverage`. In the case the user does not provide an installation the current logic of falling back to rustc llvm-tools-preview can be used by default.

`grcov` already provides this option as [`llvm-path`](https://github.com/mozilla/grcov?tab=readme-ov-file#man-grcov)

jonboh:
This allows the usage of an already installed `llvm-profdata` binary.
If the user does not provide an `llvm-path` cargo-fuzz will try to use the one installed through `rustup component add llvm-tools-preview`. 
If that fails the same instructions to install `llvm-tools-preview` as before as shown.

The name of the option and the fact that a user should set it to the bin folder of an LLVM installation (not to the binary itself) follows the same convention as `grcov`.

Closes #353 

Let me know if something needs to change, or if there's a reason why we should avoid allowing the user to set its own `llvm-profdata`

### DIFF
--- a/src/options/coverage.rs
+++ b/src/options/coverage.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use crate::{
     options::{BuildOptions, FuzzDirWrapper},
     project::FuzzProject,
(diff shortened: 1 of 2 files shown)


fitzgen (approved):
LGTM, thanks!

PR #356, opened 2024-01-06
Do not list dependency packages by MetadataCommand

kdarkhan:
When following `cargo-fuzz` tutorial at
https://rust-fuzz.github.io/book/cargo-fuzz/tutorial.html, after executing `cargo fuzz init`, `fuzz/Cargo.toml` contains entry `[dependencies.matches]` instead of `[dependencies.url]`. This is caused by `MetadataCommand` listing dependency packages in addition to main packages which is not desired.

Current implementation does work as expected if result of `MetadataCommand` contains the first package as the non-dep package but that does not seem to work all the time.

### DIFF
--- a/tests/tests/main.rs
+++ b/tests/tests/main.rs
@@ -99,6 +99,40 @@ fn init_finds_parent_project() {
 }
 
 #[test]
+fn init_defines_correct_dependency() {
+    let project_name = "project_with_some_dep";
+    let project = project(project_name)
+        .file(
+            "Cargo.toml",
+            &format!(
+                r#"
+                    [workspace]
+                    [package]
+                    name = "{name}"
+                    version = "1.0.0"
+
+                    [dependencies]
+                    matches = "0.1.10"
+                "#,
+                name = project_name
+            ),
+        )
+        .build();
+    project
+        .cargo_fuzz()
+        .current_dir(project.root().join("src"))
+        .arg("init")
+        .assert()
+        .success();
+    assert!(project.fuzz_dir().is_dir());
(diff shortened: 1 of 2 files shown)


fitzgen (commented):
I don't really follow what is going on here -- to calm my nerves, would you mind adding a regression test that fails without this change and passes after it is introduced?

kdarkhan:
> I don't really follow what is going on here -- to calm my nerves, would you mind adding a regression test that fails without this change and passes after it is introduced?

@fitzgen, just pushed a commit with a failing test case that is fixed by my change.

tests/tests/main.rs
```diff
@@ -98,6 +98,40 @@ fn init_finds_parent_project() {
...
+        .cargo_fuzz()
+        .current_dir(project.root().join("src"))
+        .arg("init")
+        .assert()
+        .success();
+    assert!(project.fuzz_dir().is_dir());
+    assert!(project.fuzz_cargo_toml().is_file());
+    let cargo_toml = fs::read_to_string(project.fuzz_cargo_toml()).unwrap();
+    let expected_dependency_attrs =
+        &format!("[dependencies.{name}]\npath = \"..\"", name = project_name);
```
kdarkhan, line 131: Based on my tests, `CargoMetadata` create calls  `cargo metadata` with or without `--no-deps`.
When called without `--no-deps`, the result contains all packages including dependency packages which are sorted alphabetically. In order to create a failing example, I had to create a sample cargo project which has a name alphabetically greater than one of its dependencies (`project_with_some_dep > matches`).

fitzgen (approved):
Thanks!

PR #365, opened 2024-03-02
Improve default_target to respect current toolchain and config

### LINKED ISSUE #355: Distributed release files can only compile for musl
See: https://github.com/astral-sh/ruff/issues/9368

It seems that binstall (which pulls from GitHub since the release) installs the musl copies distributed in the latest release, but the released files can only build musl-targeting fuzzers.

taiki-e:
Fixes #355

See https://github.com/rust-fuzz/cargo-fuzz/issues/355#issuecomment-1974748776 and https://github.com/rust-fuzz/cargo-fuzz/issues/355#issuecomment-1974758107 for details.
This PR uses the latter approach, which is more accurate. However, I have also [implemented the former approach in another branch](https://github.com/rust-fuzz/cargo-fuzz/compare/main...taiki-e:cargo-fuzz:default-target-no-dep) and can update the PR to use it if needed.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,4 +1,20 @@
+use std::sync::OnceLock;
+
 /// The default target to pass to cargo, to workaround issue #11.
 pub fn default_target() -> &'static str {
-    current_platform::CURRENT_PLATFORM
+    // Use OnceLock because clap's default_value only accepts reference.
+    static DEFAULT_TARGET: OnceLock<String> = OnceLock::new();
+    DEFAULT_TARGET.get_or_init(|| {
+        let config = cargo_config2::Config::load().unwrap();
+        // Get the target specified in config.
+        let mut targets = config.build_target_for_cli(None::<&str>).unwrap();
+        if targets.len() > 1 {
+            // Config can contain multiple targets, but we don't support it: https://doc.rust-lang.org/cargo/reference/config.html#buildtarget
+            panic!("multi-target build is not supported: {targets:?}");
+        }
+        targets
+            .pop()
+            // Get the host triple if the target is not specified in config.
+            .unwrap_or_else(|| config.host_triple().unwrap().to_owned())
+    })
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,14 +65,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "cargo-config2"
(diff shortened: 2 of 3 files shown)


src/utils.rs
```diff
@@ -1,4 +1,20 @@
...
+        // Get the target specified in config.
+        let mut targets = config.build_target_for_cli(None::<&str>).unwrap();
+        if targets.len() > 1 {
+            // Config can contain multiple targets, but we don't support it: https://doc.rust-lang.org/cargo/reference/config.html#buildtarget
+            panic!("multi-target build is not supported: {targets:?}");
+        }
+        targets
+            .pop()
+            // Get the host triple if the target is not specified in config.
+            .unwrap_or_else(|| config.host_triple().unwrap().to_owned())
```
fitzgen, line 18: I think this would be cleaner as `match targets.len() { ... }`.

src/utils.rs
```diff
@@ -1,4 +1,20 @@
...
-    current_platform::CURRENT_PLATFORM
+    // Use OnceLock because clap's default_value only accepts reference.
+    static DEFAULT_TARGET: OnceLock<String> = OnceLock::new();
+    DEFAULT_TARGET.get_or_init(|| {
+        let config = cargo_config2::Config::load().unwrap();
+        // Get the target specified in config.
+        let mut targets = config.build_target_for_cli(None::<&str>).unwrap();
+        if targets.len() > 1 {
+            // Config can contain multiple targets, but we don't support it: https://doc.rust-lang.org/cargo/reference/config.html#buildtarget
+            panic!("multi-target build is not supported: {targets:?}");
```
fitzgen, line 13: Let's return a result from this function and properly propagate errors rather than panicking here.

fitzgen (commented):
Can you report the timings for `cargo clean && time cargo build` on `main` and on this `feature` branch?

kaleidawave:
Tested the time:
```
# Old
real	0m22.610s
user	0m36.292s
sys	0m4.760s

# New
real	0m26.355s
user	0m44.987s
sys	0m5.021s
```

xtqqczze:
@taiki-e do you intend to continue with this PR?

taiki-e:
At this point, I don't have the bandwidth to work on this. Feel free to fork this PR.

PR #369, opened 2024-03-30
Annotate `coverage`, `artifacts` and `corpus` dirs  with CACHEDIR.TAG

totikom:
[CACHEDIR.TAG](https://bford.info/cachedir/) is a widely adopted way to mark directories, that should be excluded from backups, archives and so on.
IMO, `coverage`, `artifacts` and `corpus` directories should not be backed up, so this change is beneficial for setups with automatic backups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Released YYYY-MM-DD.
 
 ### Added
+- creation of `CACHEDIR.TAG` inside `artifacts`, `corpus` and `coverage` directories
 
 * TODO (or remove section if none)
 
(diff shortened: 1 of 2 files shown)


totikom:
CI is failing because of this two tests:
image
But on my machine they are also failing on `main` branch. Can someone verify, if everything is ok with this tests on main?

Shnatsel:
Artifacts are especially valuable and have very low size (~kilobytes), so I see no reason to exclude them from backups.

I also tend to back up my corpora because they took days of compute to acquire, while only taking up single-digit numbers of megabytes even for large corpora (1000+ files).

If you are looking to reduce the size of backups, a much better way to approach it would be excluding the `target` directory, which is both reproducible without investing days of compute, and is [multiple orders of magnitude larger](https://redd.it/1bms2gj).

Since this change is not clearly beneficial, and none of the maintainers can dogfood it and notice if something breaks, I'm going to go ahead and close this.

PR #376, opened 2024-07-12
Prepare for stabilization of sanitizers

Shnatsel:
The stabilization of sanitizers is expected to remove the `-Zsanitizer` flag we rely on: https://github.com/rust-lang/rust/pull/123617

This PR adds the logic to handle the compiler options required on newer rustc versions, both stable and nightly.

TODO: wait until sanitizers are actually stabilized, and fill in the correct rustc version on which they were stabilized. Hence the draft status - the code is complete otherwise.

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -1,5 +1,5 @@
 use crate::options::{self, BuildMode, BuildOptions, Sanitizer};
-use crate::rustc_version::{is_nightly, rust_version_string, RustVersion};
+use crate::rustc_version::RustVersion;
 use crate::utils::default_target;
 use anyhow::{anyhow, bail, Context, Result};
 use cargo_metadata::MetadataCommand;
@@ -7,7 +7,6 @@ use std::collections::HashSet;
 use std::io::Read;
 use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 use std::{
     env, ffi, fs,
     process::{Command, Stdio},
@@ -192,9 +191,7 @@ impl FuzzProject {
         if !matches!(build.sanitizer, Sanitizer::None) {
             // Select the appropriate sanitizer flag for the given rustc version:
             // either -Zsanitizer or -Csanitizer.
-            let rust_version_string = rust_version_string()?;
-            let rust_version =
-                RustVersion::from_str(&rust_version_string).map_err(|e| anyhow::anyhow!(e))?;
+            let rust_version = RustVersion::discover()?;
             let sanitizer_flag = match rust_version.has_sanitizers_on_stable() {
                 true => "-Csanitizer",
                 false => "-Zsanitizer",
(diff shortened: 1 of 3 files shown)


fitzgen:
Let me know when this is ready for review.

Shnatsel:
It is ready for review now!

The only remaining item is to fill in the actual version when the sanitizers are stabilized, which is not decided by the Rust project yet.

fitzgen:
Cool, thanks. I'll take a look on Monday.

src/project.rs
```diff
@@ -187,17 +189,39 @@ impl FuzzProject {
...
-            Sanitizer::Memory => {
-                // Memory sanitizer requires more flags to function than others:
-                // https://doc.rust-lang.org/unstable-book/compiler-flags/sanitizer.html#memorysanitizer
-                rustflags.push_str(" -Zsanitizer=memory -Zsanitizer-memory-track-origins")
+        if !matches!(build.sanitizer, Sanitizer::None) {
+            // Select the appropriate sanitizer flag for the given rustc version:
+            // either -Zsanitizer or -Csanitizer.
+            let rust_version_string = rust_version_string()?;
+            let rust_version =
+                RustVersion::from_str(&rust_version_string).map_err(|e| anyhow::anyhow!(e))?;
```
fitzgen, line 197: Nitpick: seems like these two steps could be collapsed into a single, fallible `RustVersion` constructor:

```rust
impl RustVersion {
    pub fn discover() -> anyhow::Result<Self> {
        let version_string = rust_version_string()?;
        let me = Self::from_str(&version_string)?;
        Ok(me)
    }
}
```

No need to make API-consumers do two things when they could do one; just an opportunity for them to mess things up in between or not connect the two interfaces properly.

src/rustc_version.rs
```diff
@@ -0,0 +1,129 @@
...
+///
+/// Stabilization PR with more info: <https://github.com/rust-lang/rust/pull/123617>
+impl RustVersion {
+    pub fn has_sanitizers_on_stable(&self) -> bool {
+        // TODO: the release that stabilizes sanitizers is not currently known.
+        // This value is a PLACEHOLDER.
+        let release_that_stabilized_sanitizers = RustVersion {
+            major: 1,
+            minor: 85,
+        };
```
fitzgen, line 79: We could land this today with

```suggestion
        // TODO: the release that stabilizes sanitizers is not currently known.
        // This value is a PLACEHOLDER.
        let release_that_stabilized_sanitizers = RustVersion {
            major: u32::MAX,
            minor: u32::MAX,
        };
```

or simply

```suggestion
        return false;
```
and then not have to wait until we know the exact version. Once we do know the exact version, it would be a small follow up PR.

fitzgen (commented):
Thanks! Couple small comments below.

src/project.rs
Shnatsel: The reason these are two separate steps is that we need the version string too, to detect whether the compiler is nightly or not.

I considered adding the version string to the `RustVersion` struct, but that would make it no longer `Copy`, and neither API looked obviously better than the other.

If you want me to encapsulate everything, I guess I could turn `RustVersion` into this:

```rust
pub struct RustVersion {
    pub major: u32,
    pub minor: u32,
    // we don't care about the patch version and it's a bit of a pain to parse
    nightly: bool,
}
```

and with the nightly flag being resolved in `RustVersion` we don't need to expose the string outside the module at all.

Shnatsel:
@fitzgen Thanks for the review! All good points, thank you!

I think I've addressed everything, so I'd appreciate you taking another look.

fitzgen (approved):
Looks good, sorry I forgot to come back to this!

PR #380, opened 2024-08-02
Set LLVM Flag to Disable cmov (select instr) generation

DanBlackwell:
Hi, I work on fuzzing and I've found this flag to be useful as it breaks up conditional checks, giving coverage feedback after each check in conditionals with `&&`s (by disabling generation of `cmov` instructions - aka `SelectInst` in LLVM API) . As you can imagine, having some checkpoints along the way helps the fuzzer break through complex checks. For example:

```
if x > 5 && y < 6 && z > 9 {
  // coverage feedback 1 set here
  do_stuff();
}
```

becomes equivalent to this:

```
if x > 5 {
  // coverage feedback 1 set here
  if y < 6 {
    // coverage feedback 2 set here
    if z > 9 {
      // coverage feedback 3 set here
      do_stuff();
    }
  }
}
```

This link shows an example in godbolt (notice how with this flag there are 3 coverage feedback entries in the generated assembly): [example](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:4,endLineNumber:15,positionColumn:4,positionLineNumber:15,selectionStartColumn:4,selectionStartLineNumber:15,startColumn:4,startLineNumber:15),source:'//+Type+your+code+here,+or+load+an+example.%0A%0A//+As+of+Rust+1.75,+small+functions+are+automatically%0A//+marked+as+%60%23%5Binline%5D%60+so+they+will+not+show+up+in%0A//+the+output+when+compiling+with+optimisations.+Use%0A//+%60%23%5Bno_mangle%5D%60+or+%60%23%5Binline(never)%5D%60+to+work+around%0A//+this+issue.%0A//+See+https://github.com/compiler-explorer/compiler-explorer/issues/5939%0A%23%5Bno_mangle%5D%0Apub+fn+square(x:+i32,+y:+i32,+z:+i32)+-%3E+i32+%7B%0A++if+x+%3E+5+%26%26+y+%3C+4+%26%26+z+%3E+9+%7B%0A++++18%0A++%7D+else+%7B%0A++++12%0A++%7D%0A%7D%0A%0A//+If+you+use+%60main()%60,+declare+it+as+%60pub%60+to+see+it+in+the+output:%0A//+pub+fn+main()+%7B+...+%7D%0A'),l:'5',n:'1',o:'Rust+source+%231',t:'0')),k:32.342454394693206,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1800,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:rust,libs:!(),options:'-C+opt-level%3D2+-Cpasses%3Dsancov-module+-Cllvm-args%3D-sanitizer-coverage-level%3D4+-Cllvm-args%3D-sanitizer-coverage-inline-8bit-counters+-Cllvm-args%3D-sanitizer-coverage-pc-table+',overrides:!(),selection:(endColumn:35,endLineNumber:19,positionColumn:35,positionLineNumber:19,selectionStartColumn:35,selectionStartLineNumber:19,startColumn:35,startLineNumber:19),source:1),l:'5',n:'0',o:'+rustc+1.80.0+(Editor+%231)',t:'0')),k:34.32421227197347,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1800,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-C+opt-level%3D2+-Cpasses%3Dsancov-module+-Cllvm-args%3D-sanitizer-coverage-level%3D4+-Cllvm-args%3D-sanitizer-coverage-inline-8bit-counters+-Cllvm-args%3D-sanitizer-coverage-pc-table+-Cllvm-args%3D-simplifycfg-branch-fold-threshold%3D0',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+rustc+1.80.0+(Editor+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

Hope this helps!

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -138,8 +138,32 @@ pub struct BuildOptions {
     #[arg(long)]
     /// Disable transformation of if-statements into `cmov` instructions (when this
     /// happens, we get no coverage feedback for that branch). Default setting is true.
-    /// A further explanation can be found here:
-    /// https://github.com/rust-fuzz/cargo-fuzz/pull/380#issue-2445529059
+    /// This is done by setting the `-simplifycfg-branch-fold-threshold=0` LLVM arg.
+    /// 
+    /// For example, in the following program shows the default coverage feedback when 
+    /// compiled with `-Copt-level=3`:
+    /// 
+    /// mark_covered(1); // mark edge 1 as covered
+    /// let mut res = 1;
+    /// if x > 5 && y < 6 {
+    ///    res = 2;
+    /// }
+    /// 
+    /// With `disable_branch_folding` enabled, the code compiles to be equivalent to:
+    /// 
+    /// mark_covered(1);
+    /// let mut res = 1;
+    /// if x > 5 {
+    ///     mark_covered(2);
+    ///     if y < 6 {
+    ///         mark_covered(3);
+    ///         res = 2;
+    ///     }
+    /// }
+    /// 
+    /// Note, that in the second program, there are now 2 new coverage feedback points,
+    /// and the fuzzer can store an input to the corpus at each condition that it passes;
+    /// giving it a better chance of producing an input that reaches `res = 2;`.
     pub disable_branch_folding: Option<bool>,
 }
 
--- a/src/options.rs
+++ b/src/options.rs
@@ -135,7 +135,7 @@ pub struct BuildOptions {
     /// available.
     pub no_trace_compares: bool,
 
-    #[arg(long)]
+    #[arg(long, default_value = "true")]
     /// Disable transformation of if-statements into `cmov` instructions (when this
     /// happens, we get no coverage feedback for that branch). Default setting is true.
     /// A further explanation can be found here:
--- a/src/options.rs
+++ b/src/options.rs
@@ -135,12 +135,12 @@ pub struct BuildOptions {
     /// available.
     pub no_trace_compares: bool,
 
-    #[arg(long, default_value = "true")]
+    #[arg(long)]
     /// Disable transformation of if-statements into `cmov` instructions (when this
     /// happens, we get no coverage feedback for that branch). Default setting is true.
     /// A further explanation can be found here:
     /// https://github.com/rust-fuzz/cargo-fuzz/pull/380#issue-2445529059
-    pub disable_branch_folding: bool,
+    pub disable_branch_folding: Option<bool>,
 }
 
 impl stdfmt::Display for BuildOptions {
@@ -240,7 +240,7 @@ mod test {
             strip_dead_code: false,
             no_cfg_fuzzing: false,
             no_trace_compares: false,
-            disable_branch_folding: true,
+            disable_branch_folding: None,
         };
 
         let opts = vec![
--- a/src/options.rs
+++ b/src/options.rs
@@ -139,18 +139,18 @@ pub struct BuildOptions {
     /// Disable transformation of if-statements into `cmov` instructions (when this
     /// happens, we get no coverage feedback for that branch). Default setting is true.
     /// This is done by setting the `-simplifycfg-branch-fold-threshold=0` LLVM arg.
-    /// 
-    /// For example, in the following program shows the default coverage feedback when 
+    ///
+    /// For example, in the following program shows the default coverage feedback when
     /// compiled with `-Copt-level=3`:
-    /// 
+    ///
     /// mark_covered(1); // mark edge 1 as covered
     /// let mut res = 1;
     /// if x > 5 && y < 6 {
     ///    res = 2;
     /// }
-    /// 
+    ///
     /// With `disable_branch_folding` enabled, the code compiles to be equivalent to:
-    /// 
+    ///
     /// mark_covered(1);
     /// let mut res = 1;
     /// if x > 5 {
@@ -160,7 +160,7 @@ pub struct BuildOptions {
     ///         res = 2;
     ///     }
     /// }
-    /// 
+    ///
     /// Note, that in the second program, there are now 2 new coverage feedback points,
     /// and the fuzzer can store an input to the corpus at each condition that it passes;
     /// giving it a better chance of producing an input that reaches `res = 2;`.
--- a/src/options.rs
+++ b/src/options.rs
@@ -134,6 +134,13 @@ pub struct BuildOptions {
     /// the `*-trace-compares` instrumentation assumes that the instruction is
     /// available.
     pub no_trace_compares: bool,
+
+    #[arg(long)]
+    /// Disable transformation of if-statements into `cmov` instructions (when this
+    /// happens, we get no coverage feedback for that branch). Default setting is true.
+    /// A further explanation can be found here:
+    /// https://github.com/rust-fuzz/cargo-fuzz/pull/380#issue-2445529059
+    pub disable_branch_folding: bool,
 }
 
 impl stdfmt::Display for BuildOptions {
@@ -233,6 +240,7 @@ mod test {
             strip_dead_code: false,
             no_cfg_fuzzing: false,
             no_trace_compares: false,
+            disable_branch_folding: true,
         };
 
         let opts = vec![
--- a/src/project.rs
+++ b/src/project.rs
@@ -176,7 +176,7 @@ impl FuzzProject {
         }
 
         if build.disable_branch_folding {
-            rustflags.push_str("-Cllvm-args=-simplifycfg-branch-fold-threshold=0");
+            rustflags.push_str(" -Cllvm-args=-simplifycfg-branch-fold-threshold=0");
         }
 
         if !build.no_cfg_fuzzing {
--- a/src/project.rs
+++ b/src/project.rs
@@ -175,7 +175,7 @@ impl FuzzProject {
             rustflags.push_str(" -Cllvm-args=-sanitizer-coverage-trace-compares");
         }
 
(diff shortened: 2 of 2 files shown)


nagisa:
This seems like a reasonable thing to do.

DanBlackwell:
I don't have any proof that it's the simplest way to achieve it, but essentially it's just setting a flag in the `SimplifyCFGPass` 
 documented [here](https://llvm.org/doxygen/SimplifyCFG_8cpp.html#a79acde6fd121317486a8a70fbc1e0c14). It basically just forces LLVM to never make the transformation that produces the `SelectInst`s (it does that [here](https://github.com/llvm/llvm-project/blob/1b8593545316971ac3f922dcb7178623b5820003/llvm/lib/Transforms/Utils/SimplifyCFG.cpp#L3939)). I can't imagine an easier way to do this than just disabling the pass (well, short of disabling all optimisation - but that seems overkill).

Sometimes it helps, sometimes it doesn't; but from what I've seen it never hurts the fuzzer.

fitzgen (commented):
Is this something that we can set even when we aren't on nightly? https://github.com/rust-fuzz/cargo-fuzz/pull/376 is preparing us for when sanitizers are stable in rustc. Although I suspect that if this new LLVM flag is nightly-only then the rest of them are as well.

------

I'd be fine merging this if we add a CLI flag like

```
--disable-branch-folding=true|false
```

with `true` being the default. Ideally this would take an `Option<bool>` in the clap configuration and then we can set the default to whatever we want given the rest of the config and context (e.g. we might disable it by default when running on non-nightly).

Shnatsel:
Yes, passing LLVM flags directly works on stable. In fact, the godbolt example uses the stable compiler without ever passing the -Z flag for any sanitizer.

My only concern here is potential performance degradation, but it shouldn't result in additional loads or stores, so it shouldn't run into extreme slowdowns from sanitizers, and degrading performance by a single digit percentage in exchange for instrumentation that's _this_ much better seems to be worth it.

<details>
<summary>Pedantic note</summary>
The godbolt example erroneously passes `-C opt-level=2` instead of `-C opt-level=3` to the compiler. 3 is the default for Cargo release builds. 
But changing it to 3 doesn't change the resulting assembly.
</details>

Shnatsel:
As for doing this without messing with LLVM passes, our best bet is an equivalent sancov option. I couldn't find any documentation for sancov options, but at least their list is defined here: https://llvm.org/doxygen/SanitizerCoverage_8cpp_source.html

I do not see anything relating to `cmov` or `SelectInst` in the list.

DanBlackwell:
> As for doing this without messing with LLVM passes, our best bet is an equivalent sancov option. I couldn't find any documentation for sancov options, but at least their list is defined here: https://llvm.org/doxygen/SanitizerCoverage_8cpp_source.html
> 
> I do not see anything relating to `cmov` or `SelectInst` in the list.

I only discovered this by discussing this with one of the sanitizer maintainers actually - their initial suggestion was to modify SanitizerCoverage myself... seemed overkill, but AFL++ does that [here](https://github.com/AFLplusplus/AFLplusplus/blob/db23931e7c1727ddac8691a6241c97b2203ec6fc/instrumentation/SanitizerCoverageLTO.so.cc#L1769). This just seems like a much simpler option...

Shnatsel:
Yeah, that (among other things) makes AFL++ difficult to compile and in turn difficult to use. That's why I almost entirely gave up on it in favor of using `cargo fuzz`.

DanBlackwell:
hmm, this doesn't do what I want:

```
    #[arg(long, default_value = "true")]
    pub disable_branch_folding: bool,
```

I could rename it to `enable_branch_folding`, but that makes it sound like we're adding a new optimisation rather than disabling a standard one. What's the best solution here?

fitzgen:
It should be an `Option<bool>` without a `default_value`. Then, when we actually use it, we do

```rust
let disable = opts.disable_branch_folding
    .unwrap_or_else(|| compute_what_the_default_should_be());
```

DanBlackwell:
Ok, got you [7990230](https://github.com/rust-fuzz/cargo-fuzz/pull/380/commits/79902307006b163dfbe2932422005f56346206d3) has it set to default to off (i.e. don't add the compiler flag), and [83cb30b](https://github.com/rust-fuzz/cargo-fuzz/pull/380/commits/83cb30bd1430bad9878b5f643f02622064b12ded) defaults to on. Let me know which you prefer.

btw, the `run_with_coverage` test fails for me even on `main`; I'm running with `cargo +nightly test`, is there anything extra I need to do or is that test failing for everyone?

src/options.rs
```diff
@@ -134,6 +134,13 @@ pub struct BuildOptions {
     /// the `*-trace-compares` instrumentation assumes that the instruction is
     /// available.
     pub no_trace_compares: bool,
+
+    #[arg(long)]
+    /// Disable transformation of if-statements into `cmov` instructions (when this
+    /// happens, we get no coverage feedback for that branch). Default setting is true.
+    /// A further explanation can be found here:
+    /// https://github.com/rust-fuzz/cargo-fuzz/pull/380#issue-2445529059
```
fitzgen, line 142: Let's inline the explanation here. It will be more likely to be kept up to date that way, and user's won't have to chase hyperlinks.

fitzgen (commented):
LGTM with one nitpick inline below, thanks!

src/options.rs
DanBlackwell: Please see the latest commit; I've condensed it as much as I can, but it's definitely a long doc comment.

fitzgen:
Looks like there is some trailing whitespace that `cargo fmt` is complaining about: https://github.com/rust-fuzz/cargo-fuzz/actions/runs/10288807948/job/28475210232?pr=380#step:4:8

> btw, the `run_with_coverage` test fails for me even on `main`; I'm running with `cargo +nightly test`, is there anything extra I need to do or is that test failing for everyone?

Testing locally now, will get back to you in a minute. It's been a while since I ran these tests, quite possible that something on nightly changed and broke them.

DanBlackwell:
Agh, I swear I'd ran it through clippy before I committed. Just fixed that, hopefully that's all for the CI failure!

fitzgen:
`run_with_coverage` passes okay for me.

I have nightly as my default in rustup, so I just did `cargo test`.

This is on `cargo 1.82.0-nightly (b5d44db1d 2024-07-26)`.

DanBlackwell:
> `run_with_coverage` passes okay for me.
> 
> I have nightly as my default in rustup, so I just did `cargo test`.
> 
> This is on `cargo 1.82.0-nightly (b5d44db1d 2024-07-26)`.

I'm on `rustc 1.82.0-nightly (60d146580 2024-08-06)` now and it still fails. I might add this is on an M-series macbook, so maybe an AArch64 difference rather than a bug.

DanBlackwell:
btw @fitzgen, clippy has some lints when I ran it; is it worth me creating a PR to fix those?

```
warning: unneeded `return` statement
   --> src/project.rs:287:13
    |
287 |             return Ok(Some(PathBuf::from(target_dir)));
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
    = note: `#[warn(clippy::needless_return)]` on by default
help: remove `return`
    |
287 -             return Ok(Some(PathBuf::from(target_dir)));
287 +             Ok(Some(PathBuf::from(target_dir)))
    |

warning: this expression creates a reference which is immediately dereferenced by the compiler
   --> src/project.rs:321:51
    |
321 |         if let Some(target_dir) = self.target_dir(&build)? {
    |                                                   ^^^^^^ help: change this to: `build`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
    = note: `#[warn(clippy::needless_borrow)]` on by default

warning: this expression creates a reference which is immediately dereferenced by the compiler
   --> src/project.rs:690:75
    |
690 |                 self.create_coverage_cmd(coverage, &coverage_out_raw_dir, &corpus.as_path())?;
    |                                                                           ^^^^^^^^^^^^^^^^^ help: change this to: `corpus.as_path()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
```

fitzgen:
I personally find that, by default, clippy is a bit too noisy for my tastes. These particular warnings all seem fine to fix.

PR #388, opened 2024-10-16
Add new build option to support Windows DLLs

### LINKED ISSUE #386: Fuzzing Windows DLL (cdylib) - Unresolved External Symbol Main
Hi cargo-fuzz devs! I am working on fuzzing a Windows DLL with cargo-fuzz and have been hitting a wall with a particular linker error. I've reproduced the problem in a small & simple Cargo project to demonstrate the issue.

## Problem Description

The issue arises when the to-be-fuzzed Cargo project defines its `crate-type` as `["cdylib"]`, specifically on Windows (it will compile into a DLL). The `Cargo.toml` looks something like this:
```toml
[package]
name = "simplelib"
version = "0.1.0"
edition = "2021"
# ...
[lib]
crate-type = ["cdylib"]
# ...
[dependencies.windows]
version = "0.58.0"
features = [ "Win32" ]
# ...
```

A very basic DLL `DllMain` function in `lib.rs` looks like this:

```rust
use windows::Win32::Foundation::*;

#[export_name = "DllMain"]
extern "system" fn dll_main(_: HINSTANCE, _: u32, _: *mut ()) -> bool
{
    true
}
```

After running `cargo fuzz init` to initialize a basic fuzzing sub-project, the command `cargo fuzz build` (or `cargo fuzz run`) will fail during building with the following linker error:

```
  = note:    Creating library C:\Users\USERNAME\dev\simplelib\fuzz\target\x86_64-pc-windows-msvc\release\deps\simplelib.dll.lib and object C:\Users\USERNAME\dev\simplelib\fuzz\target\x86_64-pc-windows-msvc\release\deps\simplelib.dll.exp
          LINK : error LNK2001: unresolved external symbol main
          C:\Users\USERNAME\dev\simplelib\fuzz\target\x86_64-pc-windows-msvc\release\deps\simplelib.dll : fatal error LNK1120: 1 unresolved externals
```

### Steps to Reproduce

1. On a Windows system (with Rust/Cargo installed), open PowerShell
2. `cargo init --lib ./simplelib`
3. (edit Cargo.toml to have the above information)
4. `cargo fuzz init`
5. `cargo fuzz build`

## Problem Analysis

I spent time digging through the various options passed to the Rust compiler by cargo-fuzz, and found [this commit](https://github.com/rust-fuzz/cargo-fuzz/commit/34ca3fbd9d7e99c1db20abf21039f17f8fa74f5c), which adds the `/include:main` linker argument for builds using the `msvc` triple. Based on the comment left with that line of code, and reading up on the [/include option](https://learn.microsoft.com/en-us/cpp/build/reference/include-force-symbol-references), I understand the intention is to force libraries compiled on Windows to include the `main` symbol, so it can be found within LibFuzzer.

I also found #281, and based on the discussion there it seems like `crate-type = ["cdylib"]` may not be supported by cargo-fuzz. If I'm thinking about this problem correctly, I understand why `cdylib` crates cause issues: DLLs/shared libraries, by nature, are loaded at runtime and don't have (don't *need*) prior knowledge of a `main` function. But, as it sits now, it appears that Cargo projects that can only build as `cdylibs` aren't able to be built by cargo-fuzz. (Please correct me if I am wrong)

## Possible Solution

Would it be possible to modify the cargo-fuzz building routine to build & instrument DLLs *without* the `main` function? This would allow the DLL to be built and instrumented in the same way as the fuzzing targets, but it would then be up to the user as to where the DLL should be dropped, so it can be loaded during fuzzing. (Since the DLL would be built with Sanitizer Coverage instrumentation, LibFuzzer should still be able to detect code coverage changes when the fuzzing target loads it in at runtime and invokes the DLL's functions.)

Perhaps something like this:

```rust
if build.triple.contains("-msvc") && !build.crate_type.ne("cdylib") {
    // The entrypoint is in the bundled libfuzzer rlib, this gets the linker to find it.
    // (Do not do this if we're building a DLL; instead let it build & be instrumented without `main`)
    rustflags.push_str(" -Clink-arg=/include:main");
}
```

I've been testing this theory locally (modifying my local copy of cargo-fuzz's code by commenting out the `rustflags.push_str(" -Clink-arg=/include:main")` when building the DLL), and it's building successfully with instrumentation. After it's built, I drop it into a location where my fuzzing target can pick it up for loading, and when I run `cargo fuzz run my_fuzz_target`, LibFuzzer appears to be quickly discovering a set of cascading if-statements and a bug I inserted inside the DLL.

cwshugg:
Hi - this PR implements a new build argument: `--no-include-main-msvc` - in order to help enable support for building and fuzzing Windows DLLs.

I recently ran into an issue when attempting to fuzz a Windows-only DLL (#386) where the `/include:main` linker argument that is added during `FuzzProject::cargo()` for MSVC-based builds gave the DLL a `main` symbol it could not resolve, which caused linking to fail:

```
LINK : error LNK2001: unresolved external symbol main
C:\....\depedency.dll : fatal error LNK1120: 1 unresolved externals
```

The DLL is, by nature, completely separate from the fuzzing targets in my cargo-fuzz repository, but I still wanted it to be a dependency listed in `fuzz/Cargo.toml`, so that it would be built and instrumented in the exact same way as the fuzzing target binaries. I found that by controlling whether or not `/include:main` is added to the `cargo build` arguments executed by cargo-fuzz, I was able to set up a "dummy" fuzzing target that provides its *own* main function:

```toml
# fuzz/Cargo.toml
# ...
[dependencies]
# the DLL is optional; only to be built with a specific feature
my_dll = { path = "..", optional = true }
# ...
[features]
build_dll = ["dep:my_dll"]
# ...
[[bin]]
name = "dummy_target_build_dll"
path = "fuzz_targets/dummy_target_build_dll.rs"
required-features = ["build_dll"]
test = false
doc = false
bench = false
# ... other binaries ...
```

```rust
// dummy_target_build_dll.rs
fn main()
{
    println!("DLL build complete! Now, copy the DLL to your desired location, `
              and use `cargo fuzz run` to run your other fuzzing targets.");
    // ... do other work, such as copying the DLL into the desired location ...
}
```

Then, by executing `cargo fuzz run --no-include-main-msvc --features=build_dll dummy_target`, the DLL would be built successfully. I could then proceed to run all other fuzzing targets, which would load in the instrumented DLL and fuzz it.

**TL;DR**: This solution allows Windows DLLs to be instrumented & fuzzed, but only if manual control over `/include:main` is allowed (hence, `--no-include-main-msvc`). This resolves #386.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -279,6 +279,7 @@ mod test {
             no_cfg_fuzzing: false,
             no_trace_compares: false,
             disable_branch_folding: None,
+            no_include_main_msvc: false,
         };
 
         let opts = vec![
--- a/src/options.rs
+++ b/src/options.rs
@@ -165,6 +165,20 @@ pub struct BuildOptions {
     /// and the fuzzer can store an input to the corpus at each condition that it passes;
     /// giving it a better chance of producing an input that reaches `res = 2;`.
     pub disable_branch_folding: Option<bool>,
+
+    #[arg(long)]
+    /// Disable the inclusion of the `/include:main` MSVC linker argument
+    ///
+    /// The purpose of `/include:main` is to force the MSVC linker to include an
+    /// external reference to the symbol `main`, such that fuzzing targets built
+    /// on Windows are able to find LibFuzzer's `main` function.
+    ///
+    /// In certain corner cases, users may prefer to *not* build with this
+    /// argument. One such example: if a user is intending to build and fuzz a
+    /// Windows DLL, they would likely choose to enable this flag, to prevent
+    /// the DLL from having an extern `main` reference added to it. (DLLs/shared
+    /// libraries should not have any reference to `main`.)
+    pub no_include_main_msvc: bool,
 }
 
 impl stdfmt::Display for BuildOptions {
--- a/src/project.rs
+++ b/src/project.rs
@@ -248,8 +248,8 @@ impl FuzzProject {
 
             // NOTE: On Windows, if the user's fuzzing targets have a dependency
             // on a local Rust DLL (with `crate-type` containing `["cdylib"]),
-            // the MSVC Linker will be unable to resolve the `main` symbol when
-            // linking the DLL. It will fail with this error:
+            // the MSVC Linker may be unable to resolve the `main` symbol when
+            // linking the DLL. It may fail with this error:
             //
             //     LINK : error LNK2001: unresolved external symbol main
             //     C:\....\depedency.dll : fatal error LNK1120: 1 unresolved externals
--- a/src/project.rs
+++ b/src/project.rs
@@ -241,28 +241,24 @@ impl FuzzProject {
         if !build.release || build.debug_assertions || build.careful_mode {
             rustflags.push_str(" -Cdebug-assertions");
         }
-        if build.triple.contains("-msvc") {
-            // The entrypoint is in the bundled libfuzzer rlib, this gets the
-            // linker to find it.
+        if build.triple.contains("-msvc") && !build.no_include_main_msvc {
+            // This forces the MSVC linker (which runs on Windows systems) to
+            // find the entry point (i.e. the `main` function) within the
+            // LibFuzzer `.rlib` file produced during the build.
+            //
+            // The `--no-include-main-msvc` argument disables the addition of
+            // this linker argument. In certain situations, a user may not want
+            // this argument included as part of the MSVC invocation.
+            //
+            // For example, if the user is attempting to build and fuzz a
+            // Windows DLL (shared library), adding `/include:main` will force
+            // the DLL to compile with an external reference to `main`.
+            // DLLs/shared libraries are designed to be built as a separate
+            // object file, intentionally left *without* knowledge of the entry
+            // point. So, by forcing a DLL to include `main` will cause linking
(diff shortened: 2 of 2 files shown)


fitzgen (approved):
Thanks!

fitzgen:
https://github.com/rust-fuzz/cargo-fuzz/actions/runs/11371974196/job/31744642568?pr=388#step:6:231

```
error[E0063]: missing field `no_include_main_msvc` in initializer of `options::BuildOptions`
Error:    --> src/options.rs:263:28
    |
263 |         let default_opts = BuildOptions {
    |                            ^^^^^^^^^^^^ missing `no_include_main_msvc`
```

cwshugg:
@fitzgen - Thanks for your patience; I was out of town the past few days and haven't had a chance to address the two PRs. Just added the fix to set that argument's default value!

cwshugg:
Hi @fitzgen - just wanted to check in and see if you've been able to give this a second look!

fitzgen:
I was on vacation for a bit, re-running CI now. Thanks for your patience!

fitzgen (approved):
Thanks!

cwshugg:
Thank you very much! Appreciate your help with getting this merged in 😄

PR #392, opened 2025-01-20
Disable stripping when building

clubby789:
The release profile now strips by default, which makes it difficult when trying to profile the fuzz target to optimise it


fitzgen (approved):
Thanks!

clubby789:
I get failures on these tests on both my branch and main 🤔 
```
failures:
    cmin
    run_without_sanitizer_with_crash
```

fitzgen (approved):

PR #402, opened 2025-02-22
Add support for more llvm instrumentations

0xdeafbeef:
Fix not working `disable_branch_folding`

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -137,6 +137,20 @@ pub struct BuildOptions {
     pub no_trace_compares: bool,
 
     #[arg(long)]
+    /// Enables `sanitizer-coverage-trace-divs` LLVM instrumentation
+    ///
+    /// When set to `true`, the compiler will instrument integer division instructions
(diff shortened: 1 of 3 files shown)


src/options.rs
```diff
@@ -165,7 +179,7 @@ pub struct BuildOptions {
     /// Note, that in the second program, there are now 2 new coverage feedback points,
     /// and the fuzzer can store an input to the corpus at each condition that it passes;
     /// giving it a better chance of producing an input that reaches `res = 2;`.
-    pub disable_branch_folding: Option<bool>,
+    pub disable_branch_folding: bool,
```
tmpolaczyk, line 182: This was an Option<bool> because the default value was "true", now the default is "false". Add this to set the default to "true":

```rust
    #[arg(long, default_value_t = true)]
```

0xdeafbeef: done

fitzgen (approved):
Thanks!

PR #408, opened 2025-04-28
Revert note about nightly rust in README

dalance:
#359 removed the note about nightly rust.
But `--sanitizer address` which is enabled by default requires `-Zsanitizer=address` which is a nightly feature.
So I think the note is necessary until merging https://github.com/rust-lang/rust/pull/123617.


fitzgen (approved):
Thanks!

PR #426, opened 2025-10-20
Add --use-libafl toggle

Carbocarde:
I recently came across this issue from 2023: https://github.com/rust-fuzz/cargo-fuzz/issues/330

Is there any interest in making libafl the default fuzzer for cargo fuzz? From what I can tell it’s faster/better than libfuzzer and is still being actively developed.

cc @manishearth @fitzgen @addisoncrump @domenukk @frewsxcv 

Stacked on #425.

### DIFF
--- a/src/options/init.rs
+++ b/src/options/init.rs
@@ -12,6 +12,10 @@ pub struct Init {
     /// Whether to create a separate workspace for fuzz targets crate
     pub fuzzing_workspace: Option<bool>,
 
+    #[arg(long, value_parser = clap::builder::BoolishValueParser::new(), default_value = "false")]
+    /// Whether to use libafl's libfuzzer_sys-compatible API for improved fuzzing performance
+    pub use_libafl: Option<bool>,
+
     #[command(flatten)]
     pub fuzz_dir_wrapper: FuzzDirWrapper,
 }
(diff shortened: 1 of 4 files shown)


fitzgen:
Adding an option seems fine by me.

I would be wary of changing the default, at least initially, and without knowing more about the project. It is very important that binaries built with `cargo fuzz` Just Work with OSS-Fuzz, for example. Not saying that they won't after this change, just that I haven't spent the time to verify that, nor has anyone else, AFAICT. And @addisoncrump's comment reinforces the idea that this should remain an optional, disabled-by-default feature for the foreseeable moment in my mind.

That said, without having actually looked at the changes yet, I'd expect the changes to be limited to `Cargo.toml` dependencies and/or `libfuzzer-sys` features enabled in the fuzz target crate, rather than changes in `cargo fuzz`. For example, you can use https://github.com/rust-fuzz/libfuzzer#linking-to-a-local-libfuzzer to link to a local `libafl++` instead of `libfuzzer`. So it seems like the necessary infrastructure to do this without any `cargo fuzz` changes already exists?

fitzgen (approved):
Ah this is just for the project initialization. Sure, seems fine by me.

addisoncrump:
> It is very important that binaries built with `cargo fuzz` Just Work with OSS-Fuzz, for example.

Yeah that's one of our long-term goals, but we want to make linkage nice and friendly first... we don't/can't test a lot of targets, yet. It could be nice to dogfood (well, kinda) fuzzing for Rust at some point.

PR #431, opened 2026-01-10
Cargo workspace support

### LINKED ISSUE #430: cargo-fuzz workspace support seems to be lacking?
Hi!

If anything this is a request for better docs.

Roughly, i have a top-level workspace, that contains a fine-grained set of members.
I would like to have the same pattern for `fuzz`, but as far as i can tell,
the `fuzz` directory must be a single package, in the sense that
1. it can't be a workspace that forces all of it's members to be fuzz-packages
2. If i add each separate fuzz package as a member of top-level workspace, then i have to build/run each one of them manually, specifying `--fuzz-dir .`
3. Since there is no way to have conditional workspace members, i'm not sure if it is possible to exclude fuzz packages from normal build

Are my observations correct? It would be nice if there could be better support for workspaces...

Thank you!

LebedevRI:
Currently, `cargo-fuzz` implements manual parser for `Cargo.toml`,
which means that only a very specific project structure is supported.
Namely, this only supports either one of:
1. Explicitly specified project directory
2. A mono-project in `"<workspace root>"/fuzz`
... where "<workspace root>" is found arbitrairly.

My biggest problem with this, is that if i want to have
a workspace-style layout for `fuzz` project,
i.e. `/fuzz/Cargo.toml` isn't a project,
but `/fuzz/foo/Cargo.toml` and `/fuzz/bar/Cargo.toml` are,
current cargo-fuzz simply doesn't support this. 

The overall solution is to stop parsing `Cargo.toml` manually,
and rely on `cargo metadata`, via existing `cargo-metadata` dep.

I've tried very hard to preserve existing functionality,
but i'm not sure how good the test coverage is.

Now, let's mention fuzz artifact dirs location (`corpus`, etc).
Currently, they will still all be located in `<workspace root>`,
but with a caveat: if the `<workspace root>/fuzz`
is part of `<workspace root>` project,
then they will no longer be stored in `fuzz` subdirectory.

Additonally, i'm wondering if this is the right default anyway,
should they be stored in, `<fuzz project root>` of each target?
Incindentally, this would restore the previous default.

Thoughts?

Fixes https://github.com/rust-fuzz/cargo-fuzz/issues/430.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,7 +128,6 @@ dependencies = [
  "rayon",
  "rustc_version",
  "tempfile",
- "toml",
 ]
 
 [[package]]
@@ -613,15 +612,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "serde_spanned"
-version = "1.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6662b5879511e06e8999a8a235d848113e942c9124f211511b16466ee2995f26"
-dependencies = [
- "serde_core",
-]
-
-[[package]]
 name = "strsim"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


LebedevRI:
Now with basic tests. The feedback i would love to hear is:
1. Is there any interest in this at all?
2. Where should the fuzz artifact dirs be located, in each project's dir, or `<workspace root>`? After thinking about this, personally i'd think they should be per-project.

LebedevRI:
(just to be explicit: this is not about different *workspaces*, this is about supporting different *projects* within a single cargo workspace. Whether or not the `<root workspace>/fuzz` is it's own workspace is a different matter, that i don't think changes the equasion much...)

LebedevRI:
I've spent some more thought on this, and ended up completely ripping out
automatic `--workspace`/`--package` insertion logic and deferring that to the user.
This is actually closer to what i initially had in mind. 

Does this look more reasonable? :) Any other suggestions?

LebedevRI:
@fitzgen kindly, poke
I'm mainly wondering if cargo workspace support is simply intentionally WONTFIX,
or there's typical review bandwidth limitations and i should just wait.
Thank you!

fitzgen:
IMO workspaces are supported just fine today (you can put multiple fuzz crates in a workspace or have them in their own workspace if you prefer) and I think making the search for the fuzz crate to operate upon more complicated isn't really worth it (today's algorithm of either taking an explicit path or walking up towards the root from the current dir is in a pretty sweet spot where it is both simple and effective).

LebedevRI:
Thanks for taking a look!

> IMO workspaces are supported just fine today (you can put multiple fuzz crates in a workspace or have them in their own workspace if you prefer)

Perhaps there is a miscommunication going on here.
Do you understand the problem i'm trying to solve here?
With current support level, how do i make cargo-fuzz build all fuzz crates in a workspace,
given that it only operates on a single fuzz crate at once?

> and I think making the search for the fuzz crate to operate upon more complicated isn't really worth it (today's algorithm of either taking an explicit path or walking up towards the root from the current dir is in a pretty sweet spot where it is both simple and effective).

LebedevRI:
@rust-fuzz @fitzgen @Manishearth who "owns" this project?
Is there some better place to clear the ongoing miscommunication?

> I think making the search for the fuzz crate to operate upon more complicated isn't really worth it (today's algorithm of either taking an explicit path or walking up towards the root from the current dir is in a pretty sweet spot where it is both simple and effective).

(I think this clearly states that cargo workspaces aren't supported by the current design...)

LebedevRI:
Ping. This isn't really inventing anything new, this is roughtly how it's done by other cargo subcommand transplants.

LebedevRI:
Gentle poke, thank you.

Manishearth:
I'm really struggling to understand the problems you are trying to solve here. I think more concrete examples and saying precisely what you expect to work might help.

I have spent ten minutes looking at this thread and I do not understand what you have been trying to say, and I do not really have the time or motivation to engage in depth here.


> but i hope that we at least agree that `cargo fuzz build`
> should actually build everything when run from the root of the cargo workspace?

I don't actually think this is a property we particularly care about?

PR #434, opened 2026-01-16
Perf/cmin parallel processing

0xalpharush:
Add a cmin strategy `speed` which batches corpora into N workers and merges workers in pairs until single, minimized corpora is produced. In my experiment, this is about 10x faster and results in about 35% more files. The minset is still correct, but we are not adding the `size` constraint that libfuzzer has (it will replace 2 files with 1 if it covers the same edges but is smaller).

Initially, I tried implementing this like `coverage` which can be run in parallel batches and aggregated at the very end. It offered some performance benefits but still had to execute about half the initial corpus serially. In order to benefit from more rounds of parallel batches and not incur too much re-execution overhead, I exploited the interrupt-recovery mechanism in libfuzzer's control file to merge batches pairwise. For example, if we merge batch A and batch B, we effectively cache the features and coverage info of A, requiring libfuzzer to only re-run B to minimize A and B. While this is only approximating the global optima wrt to size, sometimes the speed is worth it.

Because each round experiences reduction and we can cache half the inputs each round, we re-execute inputs far less than if we were not to use the caching and perform pairwise merging.

`size` strategy 
```
MERGE-OUTER: 2357 new files with 33046 new features added; 16554 new coverage edges
```

`speed` strategy
```
Merge worker 11: Merging "49" and "94"
MERGE-INNER: using the control file '/tmp/.tmp7No5Hf/49.control'
MERGE-INNER: 12551 total files; 9422 processed earlier; will process 3129 files now # Caching 
MERGE-OUTER: 2023 new files with 32403 new features added; 16547 new coverage edges
Final corpus: 3187 files
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,6 +78,7 @@ dependencies = [
  "assert_cmd",
  "cargo_metadata",
  "clap",
+ "crossbeam",
  "current_platform",
  "num_cpus",
  "predicates",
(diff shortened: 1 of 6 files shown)


stephanemagnenat:
Anyone to review this? It would be helpful for me.

Manishearth:
I don't think I have time to review large things on cargo-fuzz right now.

Issue #411, opened 2025-05-23
Request: Support for stable Rust


aep-sunlife:
Hi,

I don't like using unpinned technology stacks. How can we make some progress on getting cargo-fuzz to work with non-nightly Rust versions?

PR #440, opened 2026-04-05
Clarify stable support for --sanitizer none

ilkecan:
The previous wording implied that cargo-fuzz always requires nightly, which is too broad now that reduced `--sanitizer none` mode works on stable. Updated the docs and CLI help text to clarify that cargo-fuzz can run on stable Rust in reduced `--sanitizer none` mode.

  - keep documenting nightly as required for the default sanitizer-backed workflow
  - explain that `--sanitizer none` can work on stable
  - note that stable reduced mode does not provide sanitizer-based findings
  - call out that nightly-only options such as `--build-std`, `--careful`, and extra `-Z` flags are still not expected to work on stable
  
related: #411

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -79,7 +79,8 @@ pub struct BuildOptions {
     #[arg(long)]
     pub features: Option<String>,
 
-    /// Use a specific sanitizer
+    /// Use a specific sanitizer. `none` disables sanitizer checks and can work
+    /// on stable as long as no other nightly-only options are enabled.
     #[arg(short, long, value_enum, default_value = "address")]
     pub sanitizer: Sanitizer,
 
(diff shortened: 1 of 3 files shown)


src/options.rs
```diff
@@ -79,7 +79,8 @@ pub struct BuildOptions {
     #[arg(long)]
     pub features: Option<String>,
 
-    /// Use a specific sanitizer
+    /// Use a specific sanitizer. `none` disables sanitizer checks and can work
+    /// on stable as long as no other nightly-only options are enabled.
```
fitzgen, line 83: Small nitpick for better help message formatting:

```suggestion
    /// Use a specific sanitizer.
    ///
    /// `none` disables sanitizer checks and can work on stable as long as
    /// no other nightly-only options are enabled.
```

fitzgen (approved):
Thanks!

fitzgen:
Will merge after the nitpick is addressed, thanks.

PR #451, opened 2026-09-12
Allow --features to be passed multiple times

### LINKED ISSUE #342: The `--features` flag cannot be used multiple times
Running the following fails, while it would succeed were I using a normal `cargo run`.

```sh
cargo +nightly fuzz run my_fuzz_target --features=foo --features=bar
```

I know I can just combine this into `--features=foo,bar`, but for composability it would be nice if `cargo-fuzz` didn't require me to do so.

superbigcup325:
Fixes #342.

`cargo fuzz run my_fuzz_target --features=foo --features=bar` fails with
"the argument '--features <FEATURES>' cannot be used multiple times",
even though cargo itself accepts repeated `--features`, so wrapper
scripts that append their own `--features` have to merge them by hand.

It now matches cargo's behavior, so this works:

    cargo fuzz run my_fuzz_target --features=foo --features=bar

and each occurrence can also carry a comma-separated list:

    cargo fuzz run my_fuzz_target --features=foo,bar --features=baz

Details in the commit.

Sorry about the AI-generated description earlier — should've just
written it myself.


fitzgen:
[Please do not post LLM output verbatim in PR descriptions, comments, etc](https://noslopgrenade.com/)

tests/tests/main.rs
```diff
@@ -982,6 +982,39 @@ fn build_with_all_llvm_features() {
     assert!(a_bin.is_file(), "Not a file: {}", a_bin.display());
 }
 
+#[test]
+fn build_with_multiple_features_flags() {
+    // Regression test for #342: `--features` used to be a single-value flag,
+    // so passing it more than once failed with "cannot be used multiple
+    // times". It must match `cargo`'s behavior and accumulate.
```
fitzgen:
```suggestion
```

this comment is unnecessary

fitzgen (commented):
LGTM with below addressed

